PR #7, opened 2019-01-15
Implement logging and remove wlroots dependency

### LINKED ISSUE #1: Remove wlroots dependency
Just need to implement logging internally.

RedSoxFan:
Closes #1 

This implements a simpler version of the wlroots logger for swaylock.
With this logger, the dependency on wlroots can be dropped. This also
adds a debug flag and disables debugging output by default


emersion:
Thanks!

PR #12, opened 2019-01-16
ci: add .build.yml

emersion:


ddevault:
Can we do this on Alpine instead?

emersion:
Updated.

ddevault:
Thanks!

PR #17, opened 2019-01-16
Add a pam Meson option

emersion:
This allows CI to compile the non-PAM code too.

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -18,4 +18,7 @@ tasks:
   - build: |
       cd swaylock
       ninja -C build
-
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

ascent12:
This should be a feature option, instead of a combo.

emersion:
Yes. I've intentionally used a combo that uses the same values as a feature option. I just didn't want to bump to minimum required Meson version. But since we use already feature options for wlroots, I guess we could use those here too.

PR #20, opened 2019-01-17
Allow image file paths to contain colons

attente:
Colon characters are allowed in file paths, so if an image file named 'foo:bar.png' is passed as an --image argument, it's currently parsed as image 'bar.png' on output 'foo'.

This fix allows the caller to pass the argument ':foo:bar.png' and treat an output of '' the same as NULL. This is assuming there will never be an output named '' however...


emersion:
Build is failing, can you please rebase against the master branch?

ddevault:
Thanks!

PR #23, opened 2019-01-18
Remove VLAs.

c-edw:
Following work on Sway https://github.com/swaywm/sway/pull/3438.

Not many usages here.

### DIFF
--- a/pango.c
+++ b/pango.c
@@ -93,11 +93,11 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	char buf[max_chars];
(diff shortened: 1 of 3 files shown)


ddevault:
Can you handle the case where malloc fails?

pango.c
```diff
@@ -109,15 +113,21 @@ void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
...
 	}
 	g_object_unref(layout);
+
+	free(buf);
 }
 
 void pango_printf(cairo_t *cairo, const char *font,
 		double scale, bool markup, const char *fmt, ...) {
-	char buf[max_chars];
+	char *buf = malloc(max_chars);
```
emersion: Follow-up pull request: use `vsnprintf(NULL, 0, fmt, args)` to allocate just what we need.

c-edw: Yeah, that seems like a much better idea.

PR #24, opened 2019-01-18
Log error and exit if swaylock is suid with PAM backend.

### LINKED ISSUE #15: Abort if we have the suid bit but don't need it
If swaylock has been compiled with PAM and has suid, we should probably error out.

c-edw:
Closes #15.

I looked at how this was done on Sway. I think I know how it works - by comparing the user's ID and the effective user's ID (which might be root) we can tell if the executable has suid set.

### DIFF
--- a/pam.c
+++ b/pam.c
@@ -12,6 +12,11 @@
 static char *pw_buf = NULL;
 
 void initialize_pw_backend(void) {
+	if (getuid() != geteuid() || getgid() != getegid()) {
+		swaylock_log(LOG_ERROR,
+			"swaylock has suid but doesn't require it with PAM backend");
+		exit(EXIT_FAILURE);
+	}
 	if (!spawn_comm_child()) {
 		exit(EXIT_FAILURE);
 	}
--- a/pam.c
+++ b/pam.c
@@ -11,10 +11,11 @@
 
 static char *pw_buf = NULL;
 
-void initialize_pw_backend(void) {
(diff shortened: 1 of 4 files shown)


pam.c
```diff
@@ -12,6 +12,11 @@
 static char *pw_buf = NULL;
 
 void initialize_pw_backend(void) {
+	if (getuid() != geteuid() || getgid() != getegid()) {
+		swaylock_log(LOG_ERROR,
+			"swaylock has suid but doesn't require it with PAM backend");
```
ddevault: I think this would be clearer: "swaylock is setuid, but was compiled with the PAM backend. chmod a-x $argv[0] to fix. Aborting.".

c-edw: Should I pass `argv` into `initialize_pw_backend` or is there another way?

ddevault: Yeah, I would.

c-edw: Wait, did you mean `a-s`?

ddevault: Yeah, derp

ddevault:
You can see what the CI does here:

https://builds.sr.ht/api/jobs/22545/manifest

It's probably that it builds both with and without PAM support.

c-edw:
Yep, it was failing with PAM disabled. I'll be sure to check these manifests in future. Thanks :+1:

ddevault:
Thanks!

PR #25, opened 2019-01-18
Remove pango.c (unused)

c-edw:
See review on #23.


pango.c
```diff
@@ -93,15 +93,17 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	char *buf = malloc(max_chars);
+	va_list args;
+
+	uint32_t size = vsnprintf(NULL, 0, fmt, args);
```
emersion: `vsnprintf` returns an `int`. Can you change the type of `size` to match?

pango.c
```diff
@@ -93,15 +93,17 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
...
+	char *buf = malloc(size);
```
emersion: Need to allocate one byte more, for the terminating NULL byte (see the manpage).

c-edw: I see, `vsnprintf`s return value excludes the NULL byte.

emersion: Yeah, that's not the best API ever. :S

pango.c
```diff
@@ -93,15 +93,17 @@ PangoLayout *get_pango_layout(cairo_t *cairo, const char *font,
 
 void get_text_size(cairo_t *cairo, const char *font, int *width, int *height,
 		int *baseline, double scale, bool markup, const char *fmt, ...) {
-	char *buf = malloc(max_chars);
+	va_list args;
+
+	int size = vsnprintf(NULL, 0, fmt, args);
```
emersion: `args` is undefined at this point, maybe `va_start` should be moved above?

emersion: Disclaimer: I'm not a VA expert

c-edw: Yep: ```
The va_start() macro initializes ap for subsequent use by va_arg() and va_end(), and must be called first.
```

c-edw: Why is that? `vsnprintf` doesn't mutate `args`, does it?

emersion: See `va_arg`'s manpage.

c-edw: Actually, this file doesn't even seem to be in use. Is it an artifact of moving it to a separate repo or does it need to stay?

emersion: Oh, if it's unused, please remove it.

c-edw: Sure, I'll sweep for any others while I'm at it.

emersion:
Thanks!

PR #35, opened 2019-01-25
Allow setting both background image and colour

### LINKED ISSUE #29: Cannot use --color and --image at the same time
Contrary to the documented behaviour, specifying both --color and --image will only produce the effects of the one option which was given last (via command line or config file), i.e. either
 * the image with default transparent background (e.g. if centred), or
 * the specified colour.

bendooru:
This fixes #29 (for my use case) by unconditionally drawing the background colour and afterwards the background image, if specified.

Previously, the lockscreen would display the given background image only if it was specified and the mode was not `BACKGROUND_MODE_SOLID_COLOR`. This mode was set by the `-c` option, hence overriding a given image.

However, this changes the default background colour for the `-i` option from transparent to white, so I'm not sure if this is the proper way to fix this.

### DIFF
--- a/main.c
+++ b/main.c
@@ -658,7 +658,6 @@ static int parse_options(int argc, char **argv, struct swaylock_state *state,
 		case 'c':
 			if (state) {
 				state->args.colors.background = parse_color(optarg);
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

PR #36, opened 2019-01-27
Sleep on incorrect password with shadow backend.

### LINKED ISSUE #18: Rate limit attempts when using shadow backend
Probably just add some `sleep` calls when verification fails.

c-edw:
Closes #18.

I was thinking of perhaps doubling the timeout each time an incorrect password was entered, but decided to go with a constant 2 second timeout for the sake of simplicity.

Open to other suggestions.


ddevault:
Thanks!

PR #39, opened 2019-01-29
Add option show-failed-attempts

### LINKED ISSUE #4: 'show-failed-attempts' support
i3lock features a '--show-failed-attempts' parameter that displays the number of failed login attempts on the lock screen, it would be nice if this could get implemented for swaylock as well.

bendooru:
Displays the current number of failed attempts in the center of the indicator if `show-failed-attempts` is given, just like i3lock does.

Closes #4.


ddevault:
Thank you!

PR #42, opened 2019-02-04
Support fork-free locking notification mechanism

CameronNemo:
This allows for service/task managers to be notified when the lock screen has been displayed. It is particularly useful when locking the screen before closing the lid. The daemonization option can also be used, but then you either lose supervision of the swaylock process or resort to hacks like ptrace or cgroups.

Documentation for READY_FD can be found [here](https://gitlab.com/chinstrap/ready).

It is loosely based on [s6's service readiness notification mechanism](https://skarnet.org/software/s6/notifywhenup.html), and compatible with s6.

### DIFF
--- a/main.c
+++ b/main.c
@@ -57,6 +57,30 @@ int lenient_strcmp(char *a, char *b) {
 	}
 }
 
+#ifdef ENABLE_READYFD
+static int getenv_fd(const char *const var) {
+	char *env = getenv(var);
+
+	if (!env || env[0] == '\0') {
+		return -1;
+	}
+
+	if (unsetenv(var) != 0) {
+		return -1;
+	}
+
+	errno = 0;
+	char *endptr;
+	int fd = (int)strtol(env, &endptr, 10);
+
+	if (errno != 0 || endptr[0] != '\0') {
+		return -1;
+	}
+
+	return fd;
+}
+#endif /* ENABLE_READYFD */
+
 static void daemonize(void) {
 	int fds[2];
 	if (pipe(fds) != 0) {
@@ -520,6 +544,9 @@ static int parse_options(int argc, char **argv, struct swaylock_state *state,
 		LO_LINE_CAPS_LOCK_COLOR,
 		LO_LINE_VER_COLOR,
 		LO_LINE_WRONG_COLOR,
+#ifdef ENABLE_READYFD
+		LO_READYFD,
+#endif /* ENABLE_READYFD */
 		LO_RING_COLOR,
 		LO_RING_CLEAR_COLOR,
 		LO_RING_CAPS_LOCK_COLOR,
@@ -573,6 +600,9 @@ static int parse_options(int argc, char **argv, struct swaylock_state *state,
 		{"line-caps-lock-color", required_argument, NULL, LO_LINE_CAPS_LOCK_COLOR},
 		{"line-ver-color", required_argument, NULL, LO_LINE_VER_COLOR},
 		{"line-wrong-color", required_argument, NULL, LO_LINE_WRONG_COLOR},
+#ifdef ENABLE_READYFD
+		{"readyfd", no_argument, NULL, LO_READYFD},
+#endif /* ENABLE_READYFD */
 		{"ring-color", required_argument, NULL, LO_RING_COLOR},
 		{"ring-clear-color", required_argument, NULL, LO_RING_CLEAR_COLOR},
 		{"ring-caps-lock-color", required_argument, NULL, LO_RING_CAPS_LOCK_COLOR},
@@ -673,6 +703,10 @@ static int parse_options(int argc, char **argv, struct swaylock_state *state,
 			"Use the inside color for the line between the inside and ring.\n"
 		"  -r, --line-uses-ring             "
 			"Use the ring color for the line between the inside and ring.\n"
+#ifdef ENABLE_READYFD
+		"  --readyfd                        "
+		        "Write to the file descriptor in $READY_FD after locking.\n"
+#endif /* ENABLE_READYFD */
 		"  --ring-color <color>             "
 			"Sets the color of the ring of the indicator.\n"
 		"  --ring-clear-color <color>       "
@@ -895,6 +929,18 @@ static int parse_options(int argc, char **argv, struct swaylock_state *state,
 				state->args.colors.line.wrong = parse_color(optarg);
 			}
 			break;
+#ifdef ENABLE_READYFD
+		case LO_READYFD:
+			if (state) {
+				state->args.readyfd = getenv_fd("READY_FD");
+				if (state->args.readyfd >= 0) {
+					break;
+				}
+				swaylock_log(LOG_ERROR, "Could not get READY_FD from environment");
+				exit(EXIT_FAILURE);
+			}
+			break;
+#endif /* ENABLE_READYFD */
 		case LO_RING_COLOR:
 			if (state) {
 				state->args.colors.ring.input = parse_color(optarg);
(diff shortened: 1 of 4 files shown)


ddevault:
Can you gate this behind a meson option?

CameronNemo:
Yeah sure. Not sure why that would be necessary considering there are no new dependencies but it can't hurt :)

ddevault:
I'd just like to see this get more popular among init systems and/or downstream distributions before we start shipping it for everyone.

CameronNemo:
Added the meson option, defaulting to false.

CameronNemo:
How would you feel about a similar option in sway itself, by the way?

emersion:
>How would you feel about a similar option in sway itself, by the way?

Sway users could just put `exec ready` in their config file.

main.c
```diff
@@ -57,6 +57,42 @@ int lenient_strcmp(char *a, char *b) {
 	}
 }
 
+#ifdef ENABLE_READYFD
+static void readyfd(void) {
+	int fd;
+	char *env = NULL, *endptr = NULL;
```
emersion: Please declare those as needed.

main.c
```diff
@@ -57,6 +57,42 @@ int lenient_strcmp(char *a, char *b) {
...
+
+	env = getenv("READY_FD");
+
+	if (!env || env[0] == '\0') {
```
emersion: Is it really needed to have an argument for READY_FD? Can't we just enable it if the env variable is defined?

main.c
```diff
@@ -57,6 +57,42 @@ int lenient_strcmp(char *a, char *b) {
...
+		swaylock_log(LOG_ERROR, "Could not unset READY_FD");
+		exit(1);
+	}
+
+	errno = 0;
+	fd = (int)strtol(env, &endptr, 10);
+
+	if (errno != 0 || endptr[0] != '\0') {
+		swaylock_log(LOG_ERROR, "Could not convert READY_FD to int");
+		exit(1);
```
emersion: The spec says to print a warning in this case.

main.c
```diff
@@ -57,6 +57,42 @@ int lenient_strcmp(char *a, char *b) {
...
+	}
+
+	if (write(fd, "\n", 1) != 1) {
+		swaylock_log(LOG_ERROR, "Failed to write to readyfd");
+		close(fd);
+		exit(1);
+	}
+
+	close(fd);
+	return;
```
emersion: This `return` is not needed.

main.c
CameronNemo: I could do that. I don't in this case for two reasons:

1. While I hope this does not happen, the variable could be leaked from a parent process. In such a case the contents are likely invalid or at least not intended to be used by swaylock.

2. If a user configures their service with an older version of swaylock, the supervisor will be twiddling its thumbs while it waits for a notification that will never come. This would lead to some unexpected (although not particularly nasty) behavior. A flag would make such a misconfiguration glaringly obvious (`swaylock: unrecognized option '--readyfd'`).

main.c
CameronNemo: Since the flag is explicitly requested, I think an error is more appropriate and an update to the spec is in order.

main.c
CameronNemo: I assume you mean to not initialize them as NULL? I have done so in the most recent push. I am curious why that is a good practice.

main.c
CameronNemo: removed, sometimes I can be wordy :)

main.c
emersion: No, I mean declaring them when first used.

```c
int a, b;

… // Some code

a = f();
b = g();
```

vs.

```c
… // Some code

int a = f();
int b = g();
```

CameronNemo: done

main.c
CameronNemo: Is this a blocker?

emersion: No.

CameronNemo:
Disappointed that this was not considered for the recent release. Back to patching I guess.

emersion:
I'd really prefer this to be backed by service managers.

CameronNemo:
There are two existing.

emersion:
What are they? I don't consider that s6 supports it (it supports a similar mechanism, but not READY_FD).

Isn't it possible to discuss with service manager maintainers and ask what they think about the spec?

CameronNemo:
Unfortunately no HTML documentation, but the manpage describes the support for it in [startup](https://gitlab.com/chinstrap/startup).

Do you know of anyone else who works on service managers? Lennart shared his thoughts about the spec on twitter already. Not particularly interested in interacting with him further on this subject.

emersion:
>Lennart shared his thoughts about the spec on twitter already. Not particularly interested in interacting with him further on this subject.

For reference, discussion on Twitter: https://twitter.com/pid_eins/status/1094611367938674688

(Why shouldn't we add support for NOTIFY_SOCKET instead, which is almost the same and is widely used?)

>Do you know of anyone else who works on service managers? 

https://alternativeto.net/software/systemd/

CameronNemo:
>(Why shouldn't we add support for NOTIFY_SOCKET instead, which is almost the same and is widely used?)

You certainly could. It would require linking to libsystemd, which you already optionally do. I have some criticisms of it for other use cases, but it makes some sense here.

emersion:
Makes sense.

Moving forward, I'd be interested in what OpenRC, s6 and runit maintainers think about this. Would it be possible to get in touch and ask their thoughts?

CameronNemo:
runit maintainer is gone, has been for years. I asked some of the Void members for reviews of the reference implementation. @skarnet stopped by on the cups patch https://github.com/apple/cups/pull/5507. They did not seem to be opposed to the protocol, but then again they did not explicitly comment on the protocol but instead focused on s6 itself.

CameronNemo:
Thanks for the feedback!

emersion:
I think it would be useful to send a message on the s6 mailing list and open an issue on the OpenRC repo to ask if they would be interested in merging a patch adding READY_FD support.

skarnet:
There is no need to patch s6 for READY_FD support, it will work out of the box. Just have the run script start with `env READY_FD=n ...` if `n` is the value contained in the `notification-fd` file.

I don't know about OpenRC.

@emersion: I very much recommend _against_ supporting NOTIFY_SOCKET. NOTIFY_SOCKET is much more complex (the extra features add nothing of value, those people don't understand YAGNI) and, as the name implies, it forces the notification IPC to be a _socket_, which favours a centralized architecture such as systemd and is actively hostile to decentralized architectures such as s6. Not making any assumptions on the nature of the file descriptor used for notification is much more flexible and much better. 

Note that I provide a [wrapper](https://skarnet.org/software/misc/sdnotify-wrapper.c) for people who want to use the s6 notification mechanism under systemd; whereas to my knowledge, systemd doesn't provide anything for people who want to use the NOTIFY_SOCKET mechanism under s6. So, supporting the s6 mechanism (or READY_FD, which is almost equivalent) is more generic and can still work with systemd.

emersion:
>There is no need to patch s6 for READY_FD support, it will work out of the box. Just have the run script start with env READY_FD=n ... if n is the value contained in the notification-fd file.

Is there a reason why it doesn't work out-of-the-box?

CameronNemo:
Probably because I came later, and allowing the service to dictate the fd number requires less work for the service. (No parsing env variables needed, just write to 3 or whatever).

skarnet:
It _does_ work out-of-the-box. s6 lets the service choose any notification fd it wants; you, the user, have more _choice_, so you just need to pick a number, and tell both the service (via the READY_FD variable) and the supervisor (via the `notification-fd` file) what number you have chosen.

gdamjan:
> You certainly could. It would require linking to libsystemd, 

this is actually not true. the api is trivially implementable in anything. 
look for ex. this pure-python implementation https://github.com/bb4242/sdnotify/blob/master/sdnotify/__init__.py

WhyNotHugo:
To summarize on the s6 side:

- s6 allows the administrator to specify which file descriptor the daemon will use to notify its readiness. This is done via an unsigned integer in a config file. This value is called `notification-fd`.
- An administrator can specify environment variables for services. This includes a variable `READY_FD=`.
- An administrator can configure a service in a way that `notification-fd` matches the value of `READY_FD=`.

Essentially, the implementation on this PR has no incompatibilities. It merely requires being set up correctly.

It doesn't work out of the box because `notification-fd` must be defined explicitly. That is, support for readiness notification must be explicitly opted into by the administrator. There is no approach that will work "out of the box" because the feature is disabled by default.

Source: https://skarnet.org/software/s6/notifywhenup.html
Source: https://skarnet.org/software/s6/servicedir.html

WhyNotHugo:
Here's a rebased version of this branch:

https://github.com/swaywm/swaylock/commit/0b7bc2a2af6e4d008bfaa3a6d649e391c9e56e4f

Conflicts are trivial.

emersion:
Related: #281

emersion:
Superseded by #281

Issue #44, opened 2019-02-05
Release


jubalh:
I wanted to inquire whether you plan a new release for swaylock (and swayidle).

As I understand it the last stable version of sway included these programs. But with sway 1.0 this is going to be changed and they are seperate.

If this is the case I suggest creating new releases for swaylock and swayidle when releasing sway 1.0.

Latest release of swaylock for example doesn't even build for me because of a wrong [include path](https://github.com/swaywm/swaylock/commit/f801164b9f9d83ba28e7c60b3089691e3222a7c6). I fixed this for our local package but having everthing in order when sway 1.0 gets out would be nice.

I would also suggest to set a version automatically instead of needing to pass them at build time.
So in case git is not present just use that version. If git is present use version + git hash or whatever.

ddevault:
Let me go ahead and do a release now, actually...

jubalh:
@ddevault thanks! Edited above message just before I saw your reply. Please read again regarding version.

ddevault:
I'll tag a release now under the normal release flow but if you want to change how the version is stored, you should write a patch.

ddevault:
swaylock 1.3 is now available:

https://github.com/swaywm/swaylock/releases/tag/1.3

PR #45, opened 2019-02-05
Set version in project file

jubalh:
Let's set the version in the meson file instead of declaring it outside.

In case git is installed we use the git hash as version. Instead it
isn't (like on a clean build system), let's use the version defined in
the project.

Related to https://github.com/swaywm/swaylock/issues/44

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,7 @@
 project(
 	'swaylock',
 	'c',
+	version: '1.3',
 	license: 'MIT',
 	meson_version: '>=0.48.0',
 	default_options: [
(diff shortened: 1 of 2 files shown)


emersion:
`swaylock-version` should be removed from `meson_options.txt`.

ddevault:
Will you be making similar patches for swayidle, wlroots, and sway?

jubalh:
Yes, I plan to do them after I got feedback on this one.

jubalh:
Removed swaylock-version from meson_options.txt.
If the rest is okay I will prepare the patches for the other projects.

ddevault:
:+1:

emersion:
Is CI dead?

jubalh:
Sway: https://github.com/swaywm/sway/pull/3578

This probably is not needed. Only if we want this for the future: https://github.com/swaywm/swayidle/pull/20

Seems the other projects are good.

ddevault:
Thanks!

PR #47, opened 2019-02-05
Show current xkb layout while typing

### LINKED ISSUE #34: RFE: swaylock should display the current XKB layout
For people that have different xkb layouts configured, it can be problem when trying to unlock their computer by typing in cyrillics when their password is in latin.

swaylock should display the current xkb layout, perhaps similar to how it displays when caps-lock is pressed.

bendooru:
This fixes #34, i.e. displays the current keyboard layout while typing. Since the layout strings can be moderately long, they usually don't fit into the indicator without being unreadably small.

I thus chose to display the layout at the top of the screen in a box, like this:

grim_2019-02-05-162259485332674

I also tried putting the text directly above the indicator, but I did not like the look of it.

So there are some questions:
 * Is the current position alright?
 * Is the background box necessary or should the user make sure the text is readable on their background?
 * If using the box, are the colours OK, should they be separately configurable? Is the green (resp. ring-coloured) border necessary?

Maybe your opinions differ on how this should look.

### DIFF
--- a/render.c
+++ b/render.c
@@ -86,6 +86,7 @@ void render_frame(struct swaylock_surface *surface) {
 
 		// Draw a message
 		char *text = NULL;
+		const char *layout_text = NULL;
 		char attempts[4]; // like i3lock: count no more than 999
 		set_color_for_state(cairo, state, &state->args.colors.text);
 		cairo_select_font_face(cairo, state->args.font,
@@ -115,6 +116,20 @@ void render_frame(struct swaylock_surface *surface) {
 					snprintf(attempts, sizeof(attempts), "%d", state->failed_attempts);
 					text = attempts;
 				}
+			}
+
+			xkb_layout_index_t num_layout = xkb_keymap_num_layouts(state->xkb.keymap);
+			if (state->args.show_keyboard_layout || num_layout > 1) {
+				xkb_layout_index_t curr_layout = 0;
+
+				// advance to the first active layout (if any)
+				while (curr_layout < num_layout &&
+					xkb_state_layout_index_is_active(state->xkb.state,
+						curr_layout, XKB_STATE_LAYOUT_EFFECTIVE) != 1) {
+					++curr_layout;
+				}
+				// will handle invalid index if none are active
+				layout_text = xkb_keymap_layout_get_name(state->xkb.keymap, curr_layout);
 			}
 			break;
 		default:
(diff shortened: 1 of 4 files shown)


ddevault:
How does this look directly below the indicator?

The colors should be configurable, yes. By default please make it white text, same background as the indicator, transparent border, and a somewhat smaller font.

emersion:
By default, we could only show it if there are more than 1 layout.

ddevault:
+1

bendooru:
> By default, we could only show it if there are more than 1 layout.

As in "if enabled by option _and_ more than one layout" or  "if forced by option _or_ more than one layout"?

ddevault:
>if forced by option or more than one layout

this one

bendooru:
With your suggested changes:

grim_2019-02-05-223440251070283

Note that in the first screen shot, the indicator radius was 75, whereas here it is the default 50. In both cases, I use the font size of the indicator text (i.e. indicator-radius/3.0). Should this be determined differently or perhaps configurable?

ddevault:
Add a little padding between the lock indicator and the text. Using the font size of the lock indicator sounds good to me. Thanks!

bendooru:
Done.

render.c
```diff
@@ -116,6 +117,22 @@ void render_frame(struct swaylock_surface *surface) {
 					text = attempts;
 				}
 			}
+
+			xkb_layout_index_t num_layout = xkb_keymap_num_layouts(state->xkb.keymap);
+			if (state->args.show_keyboard_layout || num_layout > 1)
+			{
```
ddevault: style: put the { on the same line

render.c
```diff
@@ -116,6 +117,22 @@ void render_frame(struct swaylock_surface *surface) {
...
+				xkb_layout_index_t curr_layout = 0;
+
+				// advance to the first active layout (if any)
+				while (curr_layout < num_layout &&
+						xkb_state_layout_index_is_active(state->xkb.state,
+							curr_layout, XKB_STATE_LAYOUT_EFFECTIVE) != 1)
+				{
```
ddevault: Here too, and only indent contiuation lines by one indent.

render.c
```diff
@@ -185,6 +202,39 @@ void render_frame(struct swaylock_surface *surface) {
...
+			cairo_font_extents(cairo, &fe);
+			// upper left coordinates for box
+			x = (buffer_width / 2) - (extents.width / 2) - box_padding;
+			y = (buffer_height / 2) + arc_radius + arc_thickness/2 +
+				box_padding; // use box_padding also as gap to indicator
+
+			// background box
+			cairo_rectangle(cairo, x, y,
+					extents.width + 2.0 * box_padding,
+					fe.height + 2.0 * box_padding);
```
ddevault: continuation line here too

render.c
```diff
@@ -185,6 +202,39 @@ void render_frame(struct swaylock_surface *surface) {
...
+			cairo_fill_preserve(cairo);
+			// border
+			cairo_set_source_u32(cairo, state->args.colors.layout_border);
+			cairo_stroke(cairo);
+			cairo_new_sub_path(cairo);
+
+			// take font extents and padding into account
+			cairo_move_to(cairo,
+					x - extents.x_bearing + box_padding,
+					y + (fe.height - fe.descent) + box_padding);
```
ddevault: ditto

ddevault:
LGTM aside from style issues

bendooru:
This should fix the style issues.

ddevault:
Can you squash these into a single commit with a commit message suitable for the changelog?

bendooru:
Done.

ddevault:
Thanks!

PR #57, opened 2019-02-26
move all *.c files in src/

gdamjan:


ddevault:
Is there a reason for this?

emersion:
I'm pretty happy with .c files in the root directory.

gdamjan:
> Is there a reason for this?

(should've added that in the PR text.)

it makes the root directory less noisy, and easier to find either source code, or non-source-code files.

ddevault:
No thanks.

gdamjan:
> No thanks.

Is there a reason for this?

ddevault:
It's more than there's *not* a good reason for this.

PR #59, opened 2019-03-12
Decouple the drawing of the background from drawing the indicator.

andersnylander:
This should severely reduce the CPU usage of swaylock when typing, and prevent 100% CPU usage when very large images are specified as background.

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -69,6 +69,7 @@ struct swaylock_state {
 	struct loop_timer *clear_password_timer;  // clears the password buffer
 	struct wl_display *display;
 	struct wl_compositor *compositor;
+	struct wl_subcompositor *subcompositor;
 	struct zwlr_layer_shell_v1 *layer_shell;
 	struct zwlr_input_inhibit_manager_v1 *input_inhibit_manager;
 	struct wl_shm *shm;
@@ -90,8 +91,11 @@ struct swaylock_surface {
 	uint32_t output_global_name;
 	struct zxdg_output_v1 *xdg_output;
 	struct wl_surface *surface;
+	struct wl_surface *child; // surface made into subsurface
+	struct wl_subsurface *subsurface;
 	struct zwlr_layer_surface_v1 *layer_surface;
 	struct pool_buffer buffers[2];
+	struct pool_buffer indicator_buffers[2];
 	struct pool_buffer *current_buffer;
 	bool frame_pending, dirty;
 	uint32_t width, height;
@@ -111,6 +115,7 @@ struct swaylock_image {
 
 void swaylock_handle_key(struct swaylock_state *state,
 		xkb_keysym_t keysym, uint32_t codepoint);
+void render_frame_background(struct swaylock_surface *surface);
 void render_frame(struct swaylock_surface *surface);
 void render_frames(struct swaylock_state *state);
 void damage_surface(struct swaylock_surface *surface);
--- a/main.c
+++ b/main.c
@@ -126,6 +126,12 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 	surface->surface = wl_compositor_create_surface(state->compositor);
 	assert(surface->surface);
 
+	surface->child = wl_compositor_create_surface(state->compositor);
+	assert(surface->child);
+	surface->subsurface = wl_subcompositor_get_subsurface(state->subcompositor, surface->child, surface->surface);
+	assert(surface->subsurface);
+	wl_subsurface_set_sync(surface->subsurface);
+
 	surface->layer_surface = zwlr_layer_shell_v1_get_layer_surface(
 			state->layer_shell, surface->surface, surface->output,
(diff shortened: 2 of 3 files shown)


emersion:
Does this play well with transparency?

We could use a subsurface for the indicator.

andersnylander:
The current implementation doesn't care about transparency at all. Since at the moment there only exists a single surface that is being drawn to, if the indicator uses the alpha channel for anything, you'll be able to see the old frame through it.

I can look into the use of a subsurface to improve this however.

emersion:
We can't merge this if it breaks transparency. Could you look into using a subsurface?

andersnylander:
The implementation has been updated to use a subsurface for rendering the indicator, and based from what I can see, transparency works flawlessly.

The only thing left to do I guess is a code review. Would you be willing to have another look at it, @emersion?

andersnylander:
Noticed that this basically breaks the --show-keyboard-layout config parameter, since it tries to draw itself below the indicator circle, just outside the subsurface. I'll create another subsurface for handle the drawing of the layout text exclusively.

render.c
```diff
@@ -64,20 +58,52 @@ void render_frame(struct swaylock_surface *surface) {
...
+	int buffer_height = buffer_diameter * 2;
+
+	surface->current_buffer = get_next_buffer(state->shm,
+			surface->indicator_buffers, buffer_diameter, buffer_height);
+	if (surface->current_buffer == NULL) {
+		return;
+	}
+
+	// Move subsurface out of visible area
+	wl_subsurface_set_position(surface->subsurface, surface->width, surface->height);
```
emersion: Hmm. The correct way to hide the subsurface is to unmap it:

```c
wl_surface_attach(surface->child, NULL);
wl_surface_commit(surface->child);
```

render.c
```diff
@@ -233,10 +259,17 @@ void render_frame(struct swaylock_surface *surface) {
...
 	}
 
-	wl_surface_set_buffer_scale(surface->surface, surface->scale);
-	wl_surface_attach(surface->surface, surface->current_buffer->buffer, 0, 0);
+	wl_surface_set_buffer_scale(surface->child, surface->scale);
+	wl_surface_attach(surface->child, surface->current_buffer->buffer, 0, 0);
+	wl_surface_damage(surface->child, 0, 0, surface->current_buffer->width, surface->current_buffer->height);
+	wl_surface_commit(surface->child);
+
 	wl_surface_damage(surface->surface, 0, 0, surface->width, surface->height);
```
emersion: No need to damage the parent if it hasn't changed.

pool-buffer.c
```diff
@@ -128,15 +128,10 @@ struct pool_buffer *get_next_buffer(struct wl_shm *shm,
...
 	}
 
-	if (buffer->width != width || buffer->height != height) {
-		destroy_buffer(buffer);
-	}
-
-	if (!buffer->buffer) {
-		if (!create_buffer(shm, buffer, width, height,
-					WL_SHM_FORMAT_ARGB8888)) {
+	destroy_buffer(buffer);
```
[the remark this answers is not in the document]
andersnylander: The reasoning for this was due to me not figuring out how to properly reset the cairo surface to its' initial state before we start drawing a new one. With the old behaviour, we were basically overwriting the old contents, which is fine when you also redraw everything else, but not so much when you only want to redraw a single element. Since the indicator can be drawn half-transparent, the result of this is that you see the previously drawn indicator below the other. Simply deleting the buffer and creating a new one -every time- seemed to me like the by far simplest solution, though I guess a better approach would be resetting cairo surface somehow, or saving a copy of an empty surface to be copied into the existing one each time. I opted for what I felt would be simplest.

render.c
andersnylander: D'oh. I should've known there was an easier way to do it.

pool-buffer.c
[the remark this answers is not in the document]
emersion: Okay. Destroying the buffer each time is rather inefficient. Instead, you can clear the surface: https://github.com/emersion/mako/blob/master/render.c#L231

main.c
```diff
@@ -163,7 +169,7 @@ static void layer_surface_configure(void *data,
 	surface->width = width;
 	surface->height = height;
 	zwlr_layer_surface_v1_ack_configure(layer_surface, serial);
-	render_frame(surface);
+	render_frame_background(surface);
```
emersion: We should probably adjust the subsurface position for the new size here.

include/swaylock.h
```diff
@@ -111,6 +115,8 @@ struct swaylock_image {
 
 void swaylock_handle_key(struct swaylock_state *state,
 		xkb_keysym_t keysym, uint32_t codepoint);
+void render_frame_background(struct swaylock_surface *surface);
+void render_frame_repaint_background(struct swaylock_surface *surface);
```
emersion: This is unused

main.c
andersnylander: render_frame(surface) re-added.

emersion:
Thanks!

PR #65, opened 2019-03-25
Allow spaces in image file path

### LINKED ISSUE #64: Can't set background to image with spaces in name
Version info (using the Arch Linux package):
```
$ swaylock --version
swaylock version  (Mar 11 2019, branch '')
$ pacman -Qo $(which swaylock)
/usr/bin/swaylock is owned by swaylock 1.3-3
```
Regardless of how I quote the filename in the shell, swaylock fails to open it if there are spaces present.

```
$ ls -l Running\ Man\ Nebula.jpg
-rw-r--r-- 1 josh josh 1872438 Mar  6 16:33 'Running Man Nebula.jpg'
$ swaylock -i Running\ Man\ Nebula.jpg 
2019-03-25 11:16:53 - [swaylock-1.3/background-image.c:31] Failed to load background image (Failed to open file “Running”: No such file or directory).
$ swaylock -i 'Running Man Nebula.jpg'
2019-03-25 11:19:04 - [swaylock-1.3/background-image.c:31] Failed to load background image (Failed to open file “Running”: No such file or directory).
```

RedSoxFan:
Fixes #64 

This allows one or more spaces to be in the image file path.

This mostly mirrors the code in sway's background subcommand

### DIFF
--- a/main.c
+++ b/main.c
@@ -374,6 +374,23 @@ static cairo_surface_t *select_image(struct swaylock_state *state,
 	return default_image;
 }
 
+static char *join_args(char **argv, int argc) {
+	assert(argc > 0);
+	int len = 0, i;
+	for (i = 0; i < argc; ++i) {
+		len += strlen(argv[i]) + 1;
+	}
+	char *res = malloc(len);
+	len = 0;
+	for (i = 0; i < argc; ++i) {
+		strcpy(res + len, argv[i]);
+		len += strlen(argv[i]);
+		res[len++] = ' ';
+	}
+	res[len - 1] = '\0';
+	return res;
+}
+
 static void load_image(char *arg, struct swaylock_state *state) {
 	// [[<output>]:]<path>
 	struct swaylock_image *image = calloc(1, sizeof(struct swaylock_image));
@@ -407,11 +424,19 @@ static void load_image(char *arg, struct swaylock_state *state) {
 		}
 	}
 
-	// Bash doesn't replace the ~ with $HOME if the output name is supplied
+	// The shell will not expand ~ to the value of $HOME when an output name is
+	// given. Also, any image paths given in the config file need to have shell
+	// expansions performed
 	wordexp_t p;
+	while (strstr(image->path, "  ")) {
+		image->path = realloc(image->path, strlen(image->path) + 2);
+		char *ptr = strstr(image->path, "  ") + 1;
+		memmove(ptr + 1, ptr, strlen(ptr) + 1);
+		*ptr = '\\';
+	}
 	if (wordexp(image->path, &p, 0) == 0) {
 		free(image->path);
-		image->path = strdup(p.we_wordv[0]);
+		image->path = join_args(p.we_wordv, p.we_wordc);
 		wordfree(&p);
 	}
 


main.c
```diff
@@ -409,11 +426,17 @@ static void load_image(char *arg, struct swaylock_state *state) {
...
+		memmove(ptr + 1, ptr, strlen(ptr) + 1);
+		*ptr = '\\';
+	}
 	if (wordexp(image->path, &p, 0) == 0) {
 		free(image->path);
-		image->path = strdup(p.we_wordv[0]);
-		wordfree(&p);
+		image->path = join_args(p.we_wordv, p.we_wordc);
 	}
+	wordfree(&p);
```
emersion: If `wordexp` failed, `p` is undefined

emersion:
I'm really not a fan of this, but I guess we don't really have a better solution. We could only do it when the output is specified though.

main.c
RedSoxFan: Fixed

RedSoxFan:
> We could only do it when the output is specified though.

I'm not sure I follow. Regardless if there is an output provided or not, spaces were being stripped

emersion:
I mean, the only reason we do `wordexp` is to expand `~` when the output is specified. We could skip `wordexp` if not output prefix is specified.

RedSoxFan:
> I mean, the only reason we do `wordexp` is to expand `~` when the output is specified. We could skip `wordexp` if not output prefix is specified.

Doesn't wordexp also allow for stuff like:
```
swaylock --image $HOME/pictures/`hostname`-lockscreen.png
swaylock --image `/path/to/random-lockscreen-image-script`

# or in the config file
image=$HOME/pictures/`hostname`-lockscreen.png
image=`/path/to/random-lockscreen-image-script`
```

emersion:
sway's `exec` will run `sh`, which will take care of this already.

RedSoxFan:
> sway's `exec` will run `sh`, which already takes care of this.

Not if you use the config file

emersion:
Gah, you're right.

(Still not a fan of double-wordexp-ing things because escaping becomes hairy -- but fixing this would become complicated and probably wouldn't be worth it)

RedSoxFan:
There could be some syntax option to use the raw path for the paths that have issues, like `[<output>:][raw:]<path>`

main.c
```diff
@@ -409,9 +426,15 @@ static void load_image(char *arg, struct swaylock_state *state) {
 
 	// Bash doesn't replace the ~ with $HOME if the output name is supplied
 	wordexp_t p;
+	while (strstr(image->path, "  ")) {
```
emersion: Hmm, sorry if I missed something, but why are we searching for double-spaces?

main.c
```diff
@@ -409,9 +426,15 @@ static void load_image(char *arg, struct swaylock_state *state) {
 
 	// Bash doesn't replace the ~ with $HOME if the output name is supplied
```
emersion: Could we update this comment to match the reality? (ie. we also need to do a wordexp for the config file)

While you're at it, could replace "bash" with "sh"

main.c
RedSoxFan: wordexp will condense multiple spaces so stuff like `swaylock --image "path  with  multiple  spaces  in  a   row"` _(and of course, it looks like GitHub is condensing them too)_ need to have the spaces escaped.

emersion: Ehh, right

main.c
RedSoxFan: Done

emersion: Have you pushed the changes?

RedSoxFan: Oh oops sorry. Apparently, I committed, but never push. Should be pushed now

emersion:
Thanks!

PR #68, opened 2019-04-03
Fix lock indicator position on hidpi screens.

ggreer:
On hidpi screens, the lock indicator was off center (slightly up and to the left). This fixes the issue. The regression was in #59.

I'll be honest, I don't actually understand this change. I just guessed it was a scale issue and the x and y passed into `wl_subsurface_set_position()` seemed like the most likely culprit.


RedSoxFan:
Thanks!

PR #73, opened 2019-04-22
Add hide keyboard layout option

### LINKED ISSUE #60: Invert --show-keyboard-layout option
Is there any possibility of an option like --show-keyboard-layout, but where it hides the keyboard layout even if there are multiple activated?

Thanks!

jeremiah-miller:
I added a `--hide-keyboard-layout` option which can also be activated with `-K`, as an inverse of the `--show-keyboard-layout` option which can be activated with `-k`. One thing I noticed when testing was that when I didn't use the hide layout option the keyboard layout text was cut off on both sides, although my code shouldn't affect  how it was rendered and I still have that issue on master without my changes. 

Fixes #60

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -54,6 +54,7 @@ struct swaylock_args {
 	bool show_caps_lock_text;
 	bool show_caps_lock_indicator;
 	bool show_keyboard_layout;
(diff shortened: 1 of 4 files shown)


ddevault:
Thanks!

PR #76, opened 2019-04-29
Submit password by pressing Ctrl+m (close #75)

### LINKED ISSUE #75: Can't use Ctrl+m to confirm password
The last screen locker I used was `xscreensaver`, in which I could confirm the password typed with `Ctrl+m`. `swaylock` only allows me to press `Enter` instead. :disappointed:

This is by far not a big deal. But why doesn't `Ctrl+m` works the same as `Enter`? It would be nice to have it working with `Ctrl+m`.

gbrlsnchs:
This allows one to submit the password by pressing Ctrl+m, which is the same behavior as by currently pressing Ctrl+d.

### DIFF
--- a/password.c
+++ b/password.c
@@ -137,6 +137,7 @@ void swaylock_handle_key(struct swaylock_state *state,
 		schedule_indicator_clear(state);
 		schedule_password_clear(state);
 		break;
+	case XKB_KEY_m: /* fallthrough */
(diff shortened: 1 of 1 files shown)


ddevault:
Thanks!

gbrlsnchs:
Np, thank you too!

PR #78, opened 2019-05-05
Add fixed font-size option for indicator

mrschyte:
This commit adds a new command line option called --font-size to specify a fixed font size in the indicator. The default value for font-size is 0, which means that the text will autoscale. Any other positive value will result in a fixed text size.

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -47,6 +47,7 @@ struct swaylock_args {
 	struct swaylock_colors colors;
 	enum background_mode mode;
 	char *font;
+	uint32_t font_size;
 	uint32_t radius;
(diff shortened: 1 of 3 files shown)


ddevault:
Thanks!

mrschyte:
Thanks for merging. In case anyone wonders; I wanted to have an indicator similar to slock with as little effort as possible. Using the fixed font size option, you can specify a big indicator with a small text which is perfect for me. 

For example using this config:

```
font=Iosevka
font-size=24
indicator-thickness=0
indicator-radius=1024
image=~/.wallpapers/wallpaper.png
```

produces the following lockscreen:
fullscreen swaylock indicator

PR #82, opened 2019-05-20
Try to unbreak on DragonFly

### LINKED ISSUE #81: -Werror breaks build on POSIXly correct systems
`poll` is an XSI extension in POSIX < 2008, so either `_POSIX_C_SOURCE` should be bumped or replaced with `_XOPEN_SOURCE`.
```c
../loop.c: In function 'loop_poll':
../loop.c:83:2: error: implicit declaration of function 'poll'; did you mean 'powl'? [-Werror=implicit-function-declaration]
  poll(loop->fds, loop->fd_length, ms);
  ^~~~
  powl
```
http://muscles.dragonflybsd.org/synth/logs/x11___swaylock.log
https://github.com/DragonFlyBSD/DragonFlyBSD/blob/v5.4.2/sys/sys/poll.h#L99-L101
https://github.com/DragonFlyBSD/DragonFlyBSD/blob/v5.4.2/sys/sys/cdefs.h#L612-L615
https://pubs.opengroup.org/onlinepubs/009695399/functions/poll.html (POSIX 2001, X/Open 6)
https://pubs.opengroup.org/onlinepubs/9699919799/functions/poll.html (POSIX 2008, X/Open 7)

jbeich:
Fixes #81. Not tested on DragonFly as my VM no longer boots: may need to update.

### DIFF
--- a/loop.c
+++ b/loop.c
@@ -1,4 +1,4 @@
-#define _POSIX_C_SOURCE 200112L
+#define _POSIX_C_SOURCE 200809L
 #include <limits.h>
 #include <string.h>
 #include <stdbool.h>
--- a/pool-buffer.c
+++ b/pool-buffer.c
@@ -1,4 +1,4 @@
-#define _POSIX_C_SOURCE 200809
+#define _POSIX_C_SOURCE 200809L
 #include <assert.h>
 #include <cairo/cairo.h>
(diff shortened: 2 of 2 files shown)


emersion:
Thanks!

Issue #88, opened 2019-06-10
Lock screen input name is truncated in and odd way


nostdm:
**Version**

    swaylock version 1.4-4-ga9b274e (Jun  9 2019, branch 'master')

**Problem**
I have a custom keyboard layout which should display as "English (Custom)" under the password wheel on the lock screen. But instead it looks like this:

    nglish (Custo

Where the beginning and end is truncated, almost like there's a pixel width limit and if the word is too long it gets truncated in this odd way.

P.S. This used not to be the case a few versions back, it would display the whole name as expected.

PR #89, opened 2019-06-13
Resizing buffer based on the size of its content

abdelq:
Since the buffer is based on the arc radius and thickness, it is often small enough so that keymap names like "French (Canada)" or "English (the divide/multiply keys toggle the layout)" are truncated.

This is more pronounced when a bigger font size is used.

The simplest solution is pretty much to use a fullscreen buffer.

Related to #88

### DIFF
--- a/render.c
+++ b/render.c
@@ -74,15 +74,19 @@ void render_frame(struct swaylock_surface *surface) {
 	int arc_radius = state->args.radius * surface->scale;
 	int arc_thickness = state->args.thickness * surface->scale;
 	int buffer_diameter = (arc_radius + arc_thickness) * 2;
-	int buffer_height = buffer_diameter * 2;
 
-	int indicator_radius = state->args.radius + state->args.thickness;
-	int subsurf_xpos = surface->width / 2 - indicator_radius;
-	int subsurf_ypos = surface->height / 2 - indicator_radius;
+	int buffer_width = surface->indicator_width;
+	int buffer_height = surface->indicator_height;
+	int new_width = buffer_diameter;
+	int new_height = buffer_diameter;
+
+	int subsurf_xpos = surface->width / 2 - buffer_width / 2;
+	int subsurf_ypos = surface->height / 2 -
+		(state->args.radius + state->args.thickness);
 	wl_subsurface_set_position(surface->subsurface, subsurf_xpos, subsurf_ypos);
 
 	surface->current_buffer = get_next_buffer(state->shm,
-			surface->indicator_buffers, buffer_diameter, buffer_height);
+			surface->indicator_buffers, buffer_width, buffer_height);
 	if (surface->current_buffer == NULL) {
 		return;
 	}
@@ -114,7 +118,7 @@ void render_frame(struct swaylock_surface *surface) {
 	if (state->args.show_indicator && state->auth_state != AUTH_STATE_IDLE) {
 		// Draw circle
 		cairo_set_line_width(cairo, arc_thickness);
-		cairo_arc(cairo, buffer_diameter / 2, buffer_diameter / 2, arc_radius,
+		cairo_arc(cairo, buffer_width / 2, buffer_diameter / 2, arc_radius,
 				0, 2 * M_PI);
 		set_color_for_state(cairo, state, &state->args.colors.inside);
 		cairo_fill_preserve(cairo);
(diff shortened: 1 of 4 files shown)


emersion:
>The simplest solution is pretty much to use a fullscreen buffer.

But then this damages the whole screen when typing. Instead, I'd prefer to resize the subsurface.

abdelq:
@emersion I agree, but how would you get the `fe.height` and `extents.width` needed to calculate the proper subsurface when `cairo_create` is called on a `get_next_buffer`?

emersion:
The way we've done it elsewhere is by rendering first to know the size, then rendering again if the size has changed:

```c
int previous_width; // width we used for the last frame
struct buffer *buffer = get_next_buffer(previous_width);
int new_width = render(previous_buffer);
if (new_width != previous_width) {
    // size changed, get a buffer with new width
    buffer = get_next_buffer(new_width);
    render(buffer);
}
// submit buffer as usual
```

It may be necessary to update `pool-buffer.c` to be able to create dummy 0-sized buffers for the very first frame. See e.g. https://github.com/emersion/mako/blob/master/pool-buffer.c#L125.

Let me know if something's unclear.

render.c
```diff
@@ -265,7 +265,6 @@ void render_frame(struct swaylock_surface *surface) {
 			// border
 			cairo_set_source_u32(cairo, state->args.colors.layout_border);
 			cairo_stroke(cairo);
-			cairo_new_sub_path(cairo);
```
abdelq: See [documentation](https://www.cairographics.org/manual/cairo-Paths.html#cairo-new-sub-path) for rationale

abdelq:
@emersion Updated the PR.

password.c
```diff
@@ -132,6 +132,8 @@ void swaylock_handle_key(struct swaylock_state *state,
 	case XKB_KEY_Alt_R:
 	case XKB_KEY_Super_L:
 	case XKB_KEY_Super_R:
+	case XKB_KEY_ISO_Prev_Group:
```
emersion: This probably doesn't belong to this pull request.

render.c
```diff
@@ -274,6 +280,19 @@ void render_frame(struct swaylock_surface *surface) {
...
+
+			new_height += fe.height + 2 * box_padding;
+			if (new_width < (extents.width + 2 * box_padding)) {
+				new_width = extents.width + 2 * box_padding;
+			}
+		}
+
+		if (buffer_width != new_width || buffer_height != new_height) {
+			destroy_buffer(surface->current_buffer);
+			surface->current_buffer->width = new_width;
```
emersion: Hijacking `current_buffer` to store the current width/height is a little strange to me. Maybe we can add some `indicator_{width,height}` fields instead?

password.c
abdelq: Well, it's certainly more related than [3db0b6](https://github.com/swaywm/swaylock/pull/89/commits/3db0b6). The point of resizing the buffer is to properly show the keymap name, which is usually changed with those keys.

emersion: When the keymap changes, a `wl_keyboard.modifier` event is triggered. This is where we should re-render, not here.

I see that you've split this PR into two commits, which seems like a good idea to me.

render.c
```diff
@@ -274,6 +280,18 @@ void render_frame(struct swaylock_surface *surface) {
 			cairo_set_source_u32(cairo, state->args.colors.layout_text);
 			cairo_show_text(cairo, layout_text);
 			cairo_new_sub_path(cairo);
+
+			new_height += fe.height + 2 * box_padding;
+			if (new_width < (extents.width + 2 * box_padding)) {
```
emersion: Style: unnecessary parentheses

render.c
```diff
@@ -274,6 +280,18 @@ void render_frame(struct swaylock_surface *surface) {
 			cairo_set_source_u32(cairo, state->args.colors.layout_text);
 			cairo_show_text(cairo, layout_text);
 			cairo_new_sub_path(cairo);
+
+			new_height += fe.height + 2 * box_padding;
```
emersion: Hmm, is this correct? The buffer will grow vertically each time we render.

(IMHO we should probably set `new_width = new_height = 0` at the beginning of the function, always computing the new size without taking into account the old one)

password.c
abdelq: Doesn't seem to be the case. You can try it out with and without that commit and see for yourself.

render.c
abdelq: Yes this is correct. The compute of the new size doesn't take into account the old. Actually, its starting value is `buffer_diameter`. If the keymap indicator is to be rendered too, we need to take it into account and add it to the height.

password.c
emersion: I do get a `wl_keyboard.modifiers` event when switching layouts. Instead of handling these keys, please damage the state in `keyboard_modifiers`.

render.c
emersion: Oh, indeed, I got confused with `buffer_width/height` vs. `buffer_diameter`

password.c
abdelq: Updated. @emersion, thank you for your patience.

PR #92, opened 2019-06-28
fix indicator position

mortie:
On my laptop's scaled 4k display, the indicator's X position was off:
shot-2x

It was also slightly off at 1x:
shot-1x

I have to admit I don't 100% understand how the original code works, so this fix might be wrong, but it seems to work on my hardware at least. It changes the calculation of the X position to be just like how the Y position is calculated.

It also adds a 2 real pixel offset, to account for what I assume is an offset caused by integer division. A better solution than fudging the numbers like that might be to do all calculations using floats, and then round (not floor) when converting to hard pixel values, but I didn't want to make such a big change.

### DIFF
--- a/render.c
+++ b/render.c
@@ -80,9 +80,10 @@ void render_frame(struct swaylock_surface *surface) {
 	int new_width = buffer_diameter;
 	int new_height = buffer_diameter;
 
-	int subsurf_xpos = surface->width / 2 - buffer_width / 2;
+	int subsurf_xpos = surface->width / 2 -
+		(state->args.radius + state->args.thickness) + 2 / surface->scale;
 	int subsurf_ypos = surface->height / 2 -
-		(state->args.radius + state->args.thickness);
+		(state->args.radius + state->args.thickness) + 2 / surface->scale;
 	wl_subsurface_set_position(surface->subsurface, subsurf_xpos, subsurf_ypos);
 
 	surface->current_buffer = get_next_buffer(state->shm,


synaptiko:
@mortie Your fix seems to be offsetting the indicator a bit to the right for scale=1.0. I reverted to the previous code and it seems better although the circle is still not perfectly centered. I think it might be related to how Cairo does rendering, see: https://www.cairographics.org/FAQ/#sharp_lines

Unfortunately I don't have 4K monitor around so I cannot figure out what was wrong initially but here's a small helper, which can be added after the section marked as `// Clear`.
```
	cairo_set_line_width (cairo, 1);
	cairo_set_source_rgb (cairo, 0, 0, 0);
	cairo_rectangle (cairo, 0.5, 0.5, buffer_width - 1, buffer_diameter - 1);
	cairo_move_to (cairo, buffer_width / 2 + 0.5, 0);
	cairo_line_to (cairo, buffer_width / 2 + 0.5, buffer_diameter);
	cairo_move_to (cairo, 0, buffer_width / 2 + 0.5);
	cairo_line_to (cairo, buffer_width, buffer_diameter / 2 + 0.5);
	cairo_stroke (cairo);
```

By doing this you should clearly see where the buffer's area is and where is the exact center of it.

Would you mind to give it another shot on your 4K display so the buffer is centered correctly with both scale=1.0 and scale=2.0?

mortie:
@synaptiko sorry for a somewhat late reply, I've been a bit busy.

Here's a screenshot, with my change, at 4k scaled 2x:
shot-2x-patched

Here's one, with my change, at 4k 1x (I changed the color of my own image's lines to makes the square show up better):
shot-1x-patched

Here's one, from before my change, at 4k scaled 2x:
shot-2x-unpatched

Here's before my change, at 4k 1x:
shot-1x-unpatched

PR #98, opened 2019-08-04
Adds option to show indicator even if idle

afmartin:
I've found multiple times I've hit a random key to make the indicator visible (thinking I can't type my password till I see it).

### DIFF
--- a/render.c
+++ b/render.c
@@ -116,7 +116,8 @@ void render_frame(struct swaylock_surface *surface) {
 	float type_indicator_border_thickness =
 		TYPE_INDICATOR_BORDER_THICKNESS * surface->scale;
 
-	if (state->args.show_indicator && state->auth_state != AUTH_STATE_IDLE) {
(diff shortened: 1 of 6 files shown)


render.c
```diff
@@ -116,7 +116,7 @@ void render_frame(struct swaylock_surface *surface) {
 	float type_indicator_border_thickness =
 		TYPE_INDICATOR_BORDER_THICKNESS * surface->scale;
 
-	if (state->args.show_indicator && state->auth_state != AUTH_STATE_IDLE) {
+	if (state->args.show_indicator && (state->auth_state != AUTH_STATE_IDLE || state->args.indicator_idle_visible)) {
```
RedSoxFan: style: can you wrap this line so it's under 80 columns?

afmartin: Ah thanks my bad, wrapped the line.

RedSoxFan:
Thanks!

PR #108, opened 2019-09-10
Fix layout indication in various cases

im-0:


seat.c
```diff
@@ -82,7 +82,7 @@ static void keyboard_modifiers(void *data, struct wl_keyboard *wl_keyboard,
 	xkb_state_update_mask(state->xkb.state,
 		mods_depressed, mods_latched, mods_locked, 0, 0, group);
 	int caps_lock = xkb_state_mod_name_is_active(state->xkb.state,
-		XKB_MOD_NAME_CAPS, XKB_STATE_MODS_LOCKED);
+		XKB_MOD_NAME_CAPS, XKB_STATE_MODS_EFFECTIVE);
```
emersion: Is there an issue with LOCKED?

im-0: No issue. I just thought that EFFECTIVE would be more logical here.

I can remove this commit if you think that this is wrong.

render.c
```diff
@@ -168,16 +168,7 @@ void render_frame(struct swaylock_surface *surface) {
 			xkb_layout_index_t num_layout = xkb_keymap_num_layouts(state->xkb.keymap);
 			if (!state->args.hide_keyboard_layout &&
 					(state->args.show_keyboard_layout || num_layout > 1)) {
```
im-0: Also, maybe I should add `&& state->xkb.layout_set` here?

Is it possible that this code will be executed before current keyboard layout is set by keyboard_modifiers()?

im-0:
Any news?

I'm using these patches without any problems, by the way.

im-0:
Seems not needed for newer version of sway/swaylock

PR #122, opened 2020-01-06
Remove FreeBSD-specific PAM configuration

### LINKED ISSUE #121: FreeBSD PAM configuration
On FreeBSD 12.1, swaylock 1.4

The pam configuration file `pam/swaylock.freebsd` gets installed to `/usr/local/etc/pam.d/swaylock` and contains `auth include passwd`.

When swaylock is run, and I enter the correct password, it fails to unlock with the error `[pam.c:102] pam_authenticate failed: invalid credentials`
If I use the same configuration as `pam/swaylock.linux`, `auth include login`, swaylock unlocks normally.

forkbomb9:
Something in FreeBSD's PAM modules changed, and
now using the "passwd" configuration file fails to unlock
Fixes #121


ddevault:
Thanks!

Issue #3, opened 2018-04-20
Support for other pam modules


hkupty:
Hi! I know this is low priority compared to the amazing work you're doing on moving sway to wlroots, but one of the things that'd hold me back from moving to sway is the lack of support for MFA on Swaylock.
I use a yubikey as a second factor of authencication and TOTP (google auth) as a fallback in case I lose the yubikey. For some reason, it seems that neither scenarios are covered or supported by swaylock.

Anyway, thanks for the good work.

Cheers,
Henry John Kupty

gdamjan:
AFAIK under Linux pam_unix uses a suid helper /usr/bin/unix_chkpwd to check the password against /etc/shadow (which is user unreadable). So pam_yubico should either implement something similar, or use files in $HOME.

OTOH, pam_google_authenticator uses $HOME/.google_authenticator, so that should just work, but swaylock needs to support the full pam conversation.

hughwilliams94:
Just to mention that the pam-u2f module seems to work fine with swaylock, which another option for mfa with yubikey.

Spass1966:
> Just to mention that the pam-u2f module seems to work fine with swaylock, which is another option for mfa with yubikey.

@hughwilliams94 How are you using this to interact with swaylock? I have a Librem Key on the way and my understanding is it supports pam-u2f.

jsm222:
> How else are you gonna authenticate against the system password on FreeBSD? There you _must_ be root to do that, even with PAM.

Yup it does not work under FreeBSD... so you cannot unlock on FreeBSD even with correct pam settings and correct password. You can verify with pamtester that pam_unix.so only is working when pamtester is running as root

crabvk:
I followed this guide [Universal 2nd Factor](https://wiki.archlinux.org/index.php/Universal_2nd_Factor)
Have the same pam config as for sudo
```
auth   sufficient   pam_u2f.so   origin=pam://<hostname>   appid=pam://<hostname>
```
and it works.
Need to write something and press Enter before push SoloKey's button though.

PavlosMelissinos:
Can confirm that swaylock works with solokey!

`/etc/pam.d/swaylock` looks like this:

```tsv
auth sufficient pam_u2f.so authfile=/etc/u2f_mappings cue
auth include    login
```
I've also added the following line to `/etc/pam.d/login`, under `account       required    pam_nologin.so`:

```tsv
auth       sufficient   pam_u2f.so authfile=/etc/u2f_mappings cue
```

[...]

PR #125, opened 2020-01-09
Re-add FreeBSD-specific configuration

forkbomb9:
Looks like on FreeBSD, using auth include login lets swaylock unlock
with any password. But using only auth include passwd does not allow
swaylock to unlock. I have to look into it, but for now I think that
blocking the login is better than unlocking with any password. (or maybe not, tell me)

Maybe something needs to change in `pam.c`, I'll check how Linux's `pam_start` and `pam_authenticate` work and compare with FreeBSD's.


ddevault:
The FreeBSD package should just add whatever PAM rules it sees fit. PAM configurations vary wildly from distro to distro and I'm tired of trying to accomodate for them.

jsm222:
Since you as I mentioned in https://github.com/swaywm/swaylock/issues/3#issuecomment-539136733 cannot use pam_unix.so as non-root a stupid workaround is the following:
cat /etc/pam.d/swaylock 
auth           sufficient      pam_ssh.so              no_warn try_first_pass
and to have a ssh-key with a passphrase in ~/.ssh
You can then unlock with the passprhase of the key. (Not your user unix login passwd)

PR #128, opened 2020-01-23
Set version to 1.6

jubalh:
It seems this gets forgotten before doing a release.
Like last time: https://github.com/swaywm/swaylock/pull/84

So this time I will try differently and set it to the (probably)
next version number :-)


ddevault:
Thanks!

PR #134, opened 2020-03-01
Add suspend shortcut

jeremejevs:
Got fed up with having to enter a password to suspend a locked machine.

Chose to not overthink things and made a quick PoC. Most bus-related things copied from swayidle. Need some feedback:

1. Is this addition welcome?
1. How do we let the user define a shortcut? Should we just provide a few simple options (XKB layout switching style) or implement arbitrary key combinations? I'm leaning towards the former, unless the latter is a solved problem.
1. Should I move D-Bus-related stuff into a separate file?

### DIFF
--- a/include/password.h
+++ b/include/password.h
@@ -1,0 +1,6 @@
+#ifndef _SWAYLOCK_PASSWORD_H
+#define _SWAYLOCK_PASSWORD_H
+
+void connect_to_bus(void);
+
+#endif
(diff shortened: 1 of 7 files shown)


emersion:
Have you considered using `bindsym --locked`?

jeremejevs:
Now I'm *really* glad I didn't invest more time... :sweat_smile: Thanks!

PR #135, opened 2020-03-05
Added description of --indicator-x/y-position to man and completions

profojak:
Current version of swaylock supports --indicator-x-position and --indicator-y-position, but description is missing in man pages.


ddevault:
Thanks!

PR #136, opened 2020-03-25
Use xkb_keymap_new_from_buffer instead of xkb_keymap_new_from_string

NilsIrl:
xkb_keymap_new_from_string just calls xkb_keymap_new_from_buffer with size as `strlen(string)`.

https://github.com/xkbcommon/libxkbcommon/blob/0f1cae0cc482575c71d29d6241115886ebf25dec/src/keymap.c#L164

According to the [spec](https://gitlab.freedesktop.org/wayland/wayland/-/blob/master/protocol/wayland.xml#L2182), MAP_PRIVATE should be used.


ddevault:
Thanks!

PR #144, opened 2020-06-15
Implement key repeat

mortie:
This change makes Swaylock respect your key repeat rate. Even though you could argue key repeat rate isn't super important in a screen locker, I think it's important that Swaylock acts how users expect software which accepts keyboard input to work.

As part of implementing key repeat rates, this patch:
* Fixes the event loop system such that a timer can be removed from another timer's callback. This is because `swaylock_handle_key` sometimes removes timers, and `swaylock_handle_key` is now called from a timer callback. I don't know if the way I did it is the cleanest possible way, but its the best way I could think of. Calling `loop_remove_fd` from within an fd callback will still crash; I decided to not to fix that without knowing if it'll ever be a problem.
* Makes Swaylock depend on the seat interface version 4 (for the repeat_info callback).
* Changes the pointer given to keyboard callbacks to be a pointer to a `swaylock_seat` instead of a pointer to a `swaylock_state`.

### DIFF
--- a/include/seat.h
+++ b/include/seat.h
@@ -1,6 +1,11 @@
 #ifndef _SWAYLOCK_SEAT_H
 #define _SWAYLOCK_SEAT_H
 #include <xkbcommon/xkbcommon.h>
+#include <stdint.h>
+#include <stdbool.h>
+
+struct loop;
+struct loop_timer;
 
 struct swaylock_xkb {
 	bool caps_lock;
@@ -14,6 +19,11 @@ struct swaylock_seat {
 	struct swaylock_state *state;
 	struct wl_pointer *pointer;
 	struct wl_keyboard *keyboard;
+	int32_t repeat_period;
+	int32_t repeat_delay;
+	uint32_t repeat_sym;
+	uint32_t repeat_codepoint;
+	struct loop_timer *repeat_timer;
 };
 
 extern const struct wl_seat_listener seat_listener;
--- a/include/seat.h
+++ b/include/seat.h
@@ -19,8 +19,8 @@ struct swaylock_seat {
 	struct swaylock_state *state;
 	struct wl_pointer *pointer;
 	struct wl_keyboard *keyboard;
-	int32_t repeat_period;
-	int32_t repeat_delay;
+	int32_t repeat_period_ms;
+	int32_t repeat_delay_ms;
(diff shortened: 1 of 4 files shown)


mortie:
I spoke with emersion in IRC, he didn't like the changes I did to the timer (I didn't either), and he suggested setting a flag on the timer. So that's what I did in my last commit; `loop_remove_timer` no longer removes the timer directly, instead it sets a `removed` flag so that the timer gets removed the next time the "Dispatch timers" loop comes around to it.

loop.c
```diff
@@ -104,12 +105,18 @@ void loop_poll(struct loop *loop) {
 		clock_gettime(CLOCK_MONOTONIC, &now);
 		struct loop_timer *timer = NULL, *tmp_timer = NULL;
 		wl_list_for_each_safe(timer, tmp_timer, &loop->timers, link) {
+			if (timer->removed) {
+				wl_list_remove(&timer->link);
+				free(timer);
```
emersion, line 110: We should add a `continue` statement here to avoid use-after-free

include/seat.h
```diff
@@ -14,6 +19,11 @@ struct swaylock_seat {
 	struct swaylock_state *state;
 	struct wl_pointer *pointer;
 	struct wl_keyboard *keyboard;
+	int32_t repeat_period;
```
emersion, line 22: Can we add a `_ms` suffix to these fields to make it clear it's milliseconds?

seat.c
```diff
@@ -56,12 +67,25 @@ static void keyboard_key(void *data, struct wl_keyboard *wl_keyboard,
 	if (key_state == WL_KEYBOARD_KEY_STATE_PRESSED) {
 		swaylock_handle_key(state, sym, codepoint);
 	}
+
+	if (seat->repeat_timer) {
+		loop_remove_timer(seat->state->eventloop, seat->repeat_timer);
+		seat->repeat_timer = NULL;
+	}
+
+	if (key_state == WL_KEYBOARD_KEY_STATE_PRESSED && seat->repeat_period >= 0) {
```
emersion, line 76: Just in case, we could replace that `seat->repeat_period >= 0` check with a `>`

loop.c
mortie: Haha, I should've caught that. Sorry

PR #147, opened 2020-07-14
Re-read shadow password for each verification

shibe2:
This brings shadow backend behavior in line with PAM backend.

### DIFF
--- a/shadow.c
+++ b/shadow.c
@@ -40,35 +40,13 @@ void initialize_pw_backend(int argc, char **argv) {
 }
 
 void run_pw_backend_child(void) {
-	/* This code runs as root */
 	struct passwd *pwent = getpwuid(getuid());
 	if (!pwent) {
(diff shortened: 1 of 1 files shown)


shibe2:
@emersion Yes, that's a kind of disadvantage. But in my opinion it's insignificant, given that we are OK with having swaylock setuid root at all.

Another way of doing it is having a separate setuid executable that checks password. This is how PAM backend works under the hood.

emersion:
Sorry, I'm going to close this. I don't see the benefit of reading the file each time.

shibe2:
>the benefit of reading the file each time

Theoretically, if the password is changed in one session and then the user switches to another session which is locked, then currently the password he needs to enter depends on the backend. With PAM backend its the new password, which is more intuitive. With shadow backend its the old password.

I see a small benefit in making both backends behave the same way.

PR #148, opened 2020-07-16
Use explicit_bzero() if available

michaelortmann:
Use explicit_bzero() if available

The 'volatile' solution may not work as expected

Quote from https://www.gnu.org/software/libc/manual/html_node/Erasing-Sensitive-Data.html:

Declaring sensitive variables as volatile will make both the above problems worse; a volatile variable will be stored in memory for its entire lifetime, and the compiler will make more copies of it than it would otherwise have. Attempting to erase a normal variable “by hand” through a volatile-qualified pointer doesn’t work at all—because the variable itself is not volatile, some compilers will ignore the qualification on the pointer and remove the erasure anyway.

### DIFF
--- a/comm.c
+++ b/comm.c
@@ -91,7 +91,7 @@ bool write_comm_request(struct swaylock_password *pw) {
 	result = true;
 
 out:
-	explicit_bzero(pw, sizeof *pw);
+	explicit_bzero(pw, sizeof(*pw));
 	return result;
 }
 
--- a/comm.c
+++ b/comm.c
@@ -91,7 +91,7 @@ bool write_comm_request(struct swaylock_password *pw) {
 	result = true;
 
 out:
-	clear_password_buffer(pw);
+	explicit_bzero(pw, sizeof *pw);
 	return result;
 }
 
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -2,6 +2,12 @@
 #define _SWAYLOCK_H
 #include <stdbool.h>
 #include <stdint.h>
+#ifdef HAVE_EXPLICIT_BZERO
+#include <string.h>
+#include <strings.h>
+#else
+void explicit_bzero(void *buf, size_t size);
+#endif
 #include <wayland-client.h>
 #include "background-image.h"
 #include "cairo.h"
@@ -133,6 +139,5 @@ void schedule_indicator_clear(struct swaylock_state *state);
 
 void initialize_pw_backend(int argc, char **argv);
 void run_pw_backend_child(void);
-void clear_buffer(char *buf, size_t size);
 
 #endif
--- a/meson.build
+++ b/meson.build
@@ -109,10 +109,7 @@ client_protos = declare_dependency(
 
 conf_data = configuration_data()
 conf_data.set10('HAVE_GDK_PIXBUF', gdk_pixbuf.found())
-
-if cc.has_function('explicit_bzero')
-	conf_data.set('HAVE_EXPLICIT_BZERO', 1)
-endif
+conf_data.set10('HAVE_EXPLICIT_BZERO', cc.has_function('explicit_bzero'))
 
 subdir('include')
 
--- a/meson.build
+++ b/meson.build
@@ -109,7 +109,10 @@ client_protos = declare_dependency(
 
 conf_data = configuration_data()
 conf_data.set10('HAVE_GDK_PIXBUF', gdk_pixbuf.found())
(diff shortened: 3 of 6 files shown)


comm.c
```diff
@@ -91,7 +91,7 @@ bool write_comm_request(struct swaylock_password *pw) {
 	result = true;
 
 out:
-	clear_password_buffer(pw);
+	explicit_bzero(pw, sizeof *pw);
```
emersion, line 94: Style: please add parentheses around the `sizeof` argument

meson.build
```diff
@@ -110,6 +110,10 @@ client_protos = declare_dependency(
 conf_data = configuration_data()
 conf_data.set10('HAVE_GDK_PIXBUF', gdk_pixbuf.found())
 
+if cc.has_function('explicit_bzero')
+	conf_data.set('HAVE_EXPLICIT_BZERO', 1)
```
emersion, line 114: Can we use `set01` here to avoid the "forgot to include conf data header" issue? (Also need to replace `#ifdef` with `#if`)

include/swaylock.h
```diff
@@ -2,6 +2,12 @@
 #define _SWAYLOCK_H
 #include <stdbool.h>
 #include <stdint.h>
+#ifdef HAVE_EXPLICIT_BZERO
+#include <string.h>
+#include <strings.h>
```
emersion, line 7: I think only `string.h` is required here.

meson.build
```diff
@@ -110,6 +110,10 @@ client_protos = declare_dependency(
 conf_data = configuration_data()
 conf_data.set10('HAVE_GDK_PIXBUF', gdk_pixbuf.found())
 
+if cc.has_function('explicit_bzero')
```
emersion, line 113: Maybe add `prefix: "#include <string.h>"`?

emersion:
>The 'volatile' solution may not work as expected

This isn't much of an issue: `volatile` isn't used for the sensitive data, instead it's used for the zeroes.

comm.c
michaelortmann: Last project i worked on insisted to do sizeof without parenthesis. anyway. changed it to your liking.

meson.build
michaelortmann: You mean set10?

meson.build
michaelortmann: Im new to meson build system. Could you add this yourself please?

meson.build
emersion: Yeah, sorry

include/swaylock.h
emersion: I see, makes sense

meson.build
emersion:
```meson
explicit_bzero_prefix = '''#include <string.h>
#include <strings.h>'''
cc.has_function('explicit_bzero', prefix: explicit_bzero_prefix)
```

meson.build
michaelortmann: done.

meson.build
michaelortmann: Doesn't work. If done like this, meson will compile `defined(explicit_bzero)` to check for that fuction, but checking for functions with defined() is wrong, it will always return False if that symbol (function) isn't (also) a macro.

michaelortmann: In my case, it is't also a macro, and thus the detection fails.

emersion: Hmm, it sounds like the issue is that `explicit_bzero` requires `GNU_SOURCE` or `BSD_SOURCE` on glibc/musl?

And on BSDs it requires _not_ setting `_POSIX_C_SOURCE`?

michaelortmann: Added the missing args : '-D_GNU_SOURCE', works under linux glibc, should also work under musl. untested unter *bsd.

emersion:
Yeah, this requires `_GNU_SOURCE` as I thought.

I'm not sure it's worth it to maintain this extra complexity for unclear benefits.

emersion:
Sorry, I'm going to close this, we prefer to avoid non-POSIX functions.

nmeum:
> Sorry, I'm going to close this, we prefer to avoid non-POSIX functions.

You cannot reliably zero memory without avoiding non-POSIX function. explicit_bzero is nowadays implemented by OpenBSD, musl libc, and glibc (i.e. it is widely available). If you rely on the volatile keyword, chances are that the zero'ing is optimized out by the compiler. For more background information refer to https://media.ccc.de/v/35c3-9788-memsad

nmeum:
Since this issue has been closed for a couple of years now, I have opened #353 with a somewhat simpler implementation in the hopes of resuming the discussion there.

PR #151, opened 2020-08-18
Support interface localisation using GNU gettext

MisterDA:
Hi!
This is a little experiment to add localisation for the interface. Only the visible strings are translated, *not* the debug and error messages, so that they remain searchable.
The localisation is done using Meson support for GNU gettext.

The german translation was done to the best of my knowledge :wink: 

Useful links:
- https://www.gnu.org/software/gettext/manual/gettext.html#Sources
- https://mesonbuild.com/Localisation.html
- https://mesonbuild.com/i18n-module.html


MisterDA:
This new version allows presenting the localized keyboard layout, using strings from [xkeyboard-config](https://www.freedesktop.org/wiki/Software/XKeyboardConfig/), which becomes a new runtime dependency.

PR #152, opened 2020-08-30
Add -D/--dpms option to power off displays when idle

tchebb:
This option makes use of the new wlr-output-power-management protocol to power off displays when swaylock is idle (i.e. when the unlock indicator, if enabled, is not shown).

I'm not hopeful of this getting merged considering swaylock's reluctance to add new features, but (as explained in the commit message and repeated below) I believe this feature provides a significant benefit to the experience that is not possible without direct integration into swaylock:

This approach provides far more reliable DPMS control than the currently recommended solution, which is to use something above swaylock (e.g. swayidle) to coordinate swaylock invocation with a separate swaymsg invocation to control DPMS. Since swayidle is purely input-based and cannot inspect swaylock's state, it will inevitably either power up the screens when it shouldn't or fail to when it should. For example, if swayidle is set to turn on the screens on "resume," it will do so even if the only input was a mouse bump, modifier key press, or something else that swaylock ignores completely.

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -8,6 +8,7 @@
 #include "pool-buffer.h"
 #include "seat.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
+#include "wlr-output-power-management-unstable-v1-client-protocol.h"
 
 enum auth_state {
 	AUTH_STATE_IDLE,
@@ -63,6 +64,7 @@ struct swaylock_args {
 	bool show_failed_attempts;
 	bool daemonize;
 	bool indicator_idle_visible;
+	bool use_dpms;
 };
 
 struct swaylock_password {
@@ -89,6 +91,7 @@ struct swaylock_state {
 	int failed_attempts;
 	bool run_display;
 	struct zxdg_output_manager_v1 *zxdg_output_manager;
+	struct zwlr_output_power_manager_v1 *zwlr_output_power_manager;
 };
 
 struct swaylock_surface {
@@ -111,6 +114,9 @@ struct swaylock_surface {
 	enum wl_output_subpixel subpixel;
 	char *output_name;
 	struct wl_list link;
+	struct zwlr_output_power_v1 *wlr_output_power;
+	enum zwlr_output_power_v1_mode power_mode;
+	bool power_mode_pending;
 };
 
 // There is exactly one swaylock_image for each -i argument
--- a/main.c
+++ b/main.c
@@ -26,6 +26,7 @@
 #include "wlr-input-inhibitor-unstable-v1-client-protocol.h"
 #include "wlr-layer-shell-unstable-v1-client-protocol.h"
 #include "xdg-output-unstable-v1-client-protocol.h"
+#include "wlr-output-power-management-unstable-v1-client-protocol.h"
 
 static uint32_t parse_color(const char *color) {
 	if (color[0] == '#') {
@@ -210,6 +211,26 @@ static const struct wl_callback_listener surface_frame_listener = {
 	.done = surface_frame_handle_done,
 };
 
+void set_dpms(struct swaylock_state *state) {
+	if (!state->args.use_dpms) {
+		return;
+	}
+
+	enum zwlr_output_power_v1_mode intended_mode = state->auth_state == AUTH_STATE_IDLE ?
+		ZWLR_OUTPUT_POWER_V1_MODE_OFF : ZWLR_OUTPUT_POWER_V1_MODE_ON;
+
+	struct swaylock_surface *surface;
+	wl_list_for_each(surface, &state->surfaces, link) {
+		if (!surface->wlr_output_power || surface->power_mode_pending ||
+				surface->power_mode == intended_mode) {
+			continue;
+		}
+
+		zwlr_output_power_v1_set_mode(surface->wlr_output_power, intended_mode);
+		surface->power_mode_pending = true;
+	}
+}
+
 void damage_surface(struct swaylock_surface *surface) {
 	surface->dirty = true;
 	if (surface->frame_pending) {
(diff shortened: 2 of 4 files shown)


emersion:
>Since swayidle is purely input-based and cannot inspect swaylock's state, it will inevitably either power up the screens when it shouldn't or fail to when it should. For example, if swayidle is set to turn on the screens on "resume," it will do so even if the only input was a mouse bump, modifier key press, or something else that swaylock ignores completely.

I don't get it.

tchebb:
Let's say that I both want my screen to be off whenever my computer is locked and want to be able to lock the screen using a hotkey instead of waiting for it to time out. Without this feature in swaylock, I'd have to do something like this (slightly modified from the example in sway's default config):
```
exec swayidle -w \
         timeout 300 'swaylock -f ; swaymsg "output * dpms off"' \
         resume 'swaymsg "output * dpms on"'

# Force an idle state to lock the screen
bindsym --release $mod+Ctrl+l exec 'pkill -x -USR1 swayidle'
```

This tells swayidle to blank the outputs and launch swaylock whenever I'm idle, and the hotkey just forces that state. The hotkey has to coordinate with swayidle instead of just running `swaylock -f ; swaymsg "output * dpms off` on its own because otherwise nothing would turn the screens back on if I pressed the hotkey and then tried to unlock within 300 seconds.

There are two problems with this configuration:
1. After I lock the screen, swayidle still tracks idle state and runs commands. New swaylock invocations will fail harmlessly if the screen is already locked, but the DPMS commands will execute as normal. This means that, if I bump my mouse, accidentally press a key, or do anything else to un-idle the session while my computer is locked, my screen turns on and shows swaylock's background color for 600 seconds. This wastes power and frankly just looks unprofessional.
2. Even with `--release` in my binding, it seems like sometimes the lock hotkey itself is enough to trigger a resume event right after it forces an idle. This means that my screen often stays on for 600 seconds right after I lock it, without me doing anything additional to trigger a resume. (Admittedly, there are probably other ways to fix this specific problem, like adding a delay before sending USR1 to wait for the keys to be released or by patching swayidle to ignore key release events, but problem 1 will remain.)

The underlying issue here is that there's no guarantee that swayidle will consider the system idle (and so keep the displays off) if and only if swaylock is running. And there's no way to construct that guarantee either, since swayidle can only observe input activity and timeouts: swaylock might start before a timeout has elapsed, and it might continue running even after there's input activity. So the states will inevitably get out of sync, meaning the screens will turn on when they shouldn't.

With this patch, the configuration is way simpler:
```
exec swayidle -w timeout 300 'swaylock -fD'
# Force an idle state to lock the screen
bindsym --release $mod+Ctrl+l exec 'pkill -x -USR1 swayidle'
```

And guarantees that the screens will be off whenever the system is locked and there's not an unlock attempt in progress.

emersion:
>New swaylock invocations will fail harmlessly if the screen is already locked, but the DPMS commands will execute as normal

```
swaylock -f && swaymsg "output * dpms off
```

>Even with --release in my binding, it seems like sometimes the lock hotkey itself is enough to trigger a resume event right after it forces an idle

I use `sleep 1` to prevent this.

tchebb:
> ```
> swaylock -f && swaymsg "output * dpms off
> ```
The issue isn't the `dpms off` command, it's the `dpms on` command. If I accidentally bump the mouse, I don't want my monitor to turn on for five minutes. In fact, the invocation you suggest will just make that situation worse as it will mean my monitor will stay on until I next unlock the system.
> > Even with --release in my binding, it seems like sometimes the lock hotkey itself is enough to trigger a resume event right after it forces an idle
> 
> I use `sleep 1` to prevent this.
Like I said, there are fixes to issue #2 (although I think the `sleep 1` fix is pretty hacky). But, again, those don't address the underlying problem of swaylock and swayidle getting out of sync.

emersion:
> If I accidentally bump the mouse, I don't want my monitor to turn on for five minutes.

But you want to be able to turn on the screen while unlocking swaylock? Not sure how that'd work.

All in all, I think I'd rather keep swayidle a generic dispatching tool. It would be nice to have a fork/new tool that provides deeper integration with DPMS.

tchebb:
I agree that it can't work if implemented in swayidle. swayidle has no way to distinguish between input that starts an unlock and other input that should be ignored. And I can't think of any way to give it that awareness short of creating a new protocol for the lockscreen to inform swayidle of its state, at which point we might as well just make the lockscreen control DPMS directly. Which is exactly what this PR does.

What's your objection to this PR? It "keep[s] swayidle a generic dispatching tool" just like you said, confining all the DPMS logic to the lockscreen. I think the lockscreen is exactly the right place to do DPMS: it controls the images on the displays, so why shouldn't it control their power state as well? And, like we've both pointed out, it's the only component that knows when something "interesting" is happening and the displays should be lit up.

Sure, in the case of swaylock, there are hacks with swayidle to do DPMS at mostly the right times, but they're not as good as what this PR provides. And we can imagine other lockscreens where using swayidle for DPMS would be totally infeasible: for example, a lockscreen for mobile that shows notifications and wants to light up the displays for a few seconds when a new one comes in (a la Android). There's no possible way that any other component than the lockscreen itself could control DPMS in that case. Which just convinces me further that the lockscreen is the right place for this logic.

emersion:
Oh, I was mistaken and thought this PR was about swayidle.

>distinguish between input that starts an unlock and other input that should be ignored

Never heard of a DE that doesn't DPMS off on any activity, even if the lockscreen ignores said activity.

>the lockscreen is exactly the right place to do DPMS

Not really, no. It may be for your use-case, but not in general. For instance users may want to DMPS without locking.

* * *

This is a hard -1 from me. swaylock has nothing to do with DPMS.

emersion:
Of course, feel free to fork/write a tool that has idle, DPMS and locking all integrated. I think that'd be great and that other people would be interested as well.

tchebb:
If you've found this PR and are interested in the functionality, I've also submitted the same change to swaylock-effects as mortie/swaylock-effects#40.

PR #154, opened 2020-10-14
Fix a potential use-after-free

mortie:
I haven't looked too deep into exactly how the buffer system works, but from what I understand, there's an opportunity for a use-after-free if a `pool_buffer` is freed without calling `destroy_buffer`, if Wayland decides to send the `buffer_release` event after the `free` takes place. I don't know if this would ever happen in practice in swaylock, but in swaylock-effects, this issue would trigger a use-after-free if an output is destroyed while a fade-in animation is running.


emersion (approved):
LGTM, thanks!

PR #155, opened 2020-10-15
Fix output-specific images when the output reappears after being disconnected

mortie:
When setting an image with `--image <output>:<path>`, the image used to fail to apply if the relevant output appears some time after swaylock executes. That's because the `zxdg_output_manager` stuff was only done from the initialization code in main, and `create_layer_surface` depends on the `output_name` property to find the right image, which is set by the `xdg_output_listener`.

One thing to note is that this patch adds a window of time where the surface exists, but hasn't been initialized. Before this patch, the event loop was blocked until the surface was ready, but now, `wl_display_roundtrip` is called _before_ the surface is initialized with `create_layer_surface`. This state is indicated with the new `surface->ready` property. From what I can tell, the only part which really needs to care is `damage_surface`, but someone more familiar with the code base should think through it.

### DIFF
--- a/main.c
+++ b/main.c
@@ -131,6 +131,7 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 				state->zxdg_output_manager, surface->output);
 		zxdg_output_v1_add_listener(
 				surface->xdg_output, &_xdg_output_listener, surface);
+		surface->events_pending += 1;
 	} else if (!has_printed_zxdg_error) {
 		swaylock_log(LOG_INFO, "Compositor does not support zxdg output "
 				"manager, images assigned to named outputs will not work");
@@ -162,7 +163,12 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 			surface->layer_surface, true);
 	zwlr_layer_surface_v1_add_listener(surface->layer_surface,
 			&layer_surface_listener, surface);
+	surface->events_pending += 1;
 
+	wl_surface_commit(surface->surface);
+}
+
+static void initially_render_surface(struct swaylock_surface *surface) {
 	if (surface_is_opaque(surface) &&
 			surface->state->args.mode != BACKGROUND_MODE_CENTER &&
 			surface->state->args.mode != BACKGROUND_MODE_FIT) {
@@ -173,7 +179,8 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 		wl_region_destroy(region);
 	}
 
-	wl_surface_commit(surface->surface);
+	render_frame_background(surface);
+	render_frame(surface);
 }
 
 static void layer_surface_configure(void *data,
@@ -185,8 +192,10 @@ static void layer_surface_configure(void *data,
 	surface->indicator_width = 1;
 	surface->indicator_height = 1;
 	zwlr_layer_surface_v1_ack_configure(layer_surface, serial);
-	render_frame_background(surface);
-	render_frame(surface);
+
+	if (--surface->events_pending == 0) {
+		initially_render_surface(surface);
+	}
 }
 
 static void layer_surface_closed(void *data,
@@ -307,9 +316,10 @@ static void handle_xdg_output_done(void *data, struct zxdg_output_v1 *output) {
 	cairo_surface_t *new_image = select_image(surface->state, surface);
 	if (new_image != surface->image) {
 		surface->image = new_image;
-		if (surface->state->run_display) {
-			damage_surface(surface);
-		}
+	}
+
+	if (--surface->events_pending == 0) {
+		initially_render_surface(surface);
 	}
 }
 
--- a/main.c
+++ b/main.c
@@ -173,8 +173,6 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 		wl_region_destroy(region);
 	}
 
-	surface->ready = true;
-
 	wl_surface_commit(surface->surface);
 }
 
@@ -227,10 +225,6 @@ static const struct wl_callback_listener surface_frame_listener = {
 };
 
 void damage_surface(struct swaylock_surface *surface) {
-	if (!surface->ready) {
-		return;
-	}
-
 	surface->dirty = true;
 	if (surface->frame_pending) {
 		return;
--- a/main.c
+++ b/main.c
@@ -357,7 +357,6 @@ static void handle_global(void *data, struct wl_registry *registry,
 				&wl_output_interface, 3);
 		surface->output_global_name = name;
 		wl_output_add_listener(surface->output, &_wl_output_listener, surface);
-
 		wl_list_insert(&state->surfaces, &surface->link);
 
 		if (state->run_display) {
--- a/main.c
+++ b/main.c
@@ -159,6 +159,8 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 		wl_region_destroy(region);
 	}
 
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -336,6 +342,18 @@ static void handle_global(void *data, struct wl_registry *registry,
 				&wl_output_interface, 3);
 		surface->output_global_name = name;
 		wl_output_add_listener(surface->output, &_wl_output_listener, surface);
+
+		if (state->zxdg_output_manager) {
+			surface->xdg_output = zxdg_output_manager_v1_get_xdg_output(
+					state->zxdg_output_manager, surface->output);
+			zxdg_output_v1_add_listener(
+					surface->xdg_output, &_xdg_output_listener, surface);
+			wl_display_roundtrip(state->display);
```
emersion, line 351: A roundtrip inside of a Wayland event handle is bad practice and can lead to lockups.

main.c
```diff
@@ -336,6 +342,18 @@ static void handle_global(void *data, struct wl_registry *registry,
...
+		} else {
+			swaylock_log(LOG_INFO, "Compositor does not support zxdg output "
```
emersion, line 353: We probably want to display this message only once, not once per output.

main.c
mortie: Hmm, but the original code does it too? https://github.com/swaywm/swaylock/blob/235b925df7e1bb82d98f1ac8c02e8f85d0a54ee9/main.c#L345

Is it fine when it's protected by the `if (state->run_display)`?

emersion: The original code does it outside of the event handler.

mortie: The original code inside the `handle_global` event handler looks like:

``` C
	} else if (strcmp(interface, wl_output_interface.name) == 0) {
		struct swaylock_surface *surface =
			calloc(1, sizeof(struct swaylock_surface));
		surface->state = state;
		surface->output = wl_registry_bind(registry, name,
				&wl_output_interface, 3);
		surface->output_global_name = name;
		wl_output_add_listener(surface->output, &_wl_output_listener, surface);
		wl_list_insert(&state->surfaces, &surface->link);

		if (state->run_display) {
			create_layer_surface(surface);
			wl_display_roundtrip(state->display);
		}
	}
```

Isn't that a `wl_display_roundtrip` in the event handler?

emersion: Oh well. Then current code is wrong.

In this particular case, I think the existing `wl_display_roundtrip` can be dropped, because there's no code that runs after and depends on some events.

mortie: Alright. I can try to move the logic out of the event handler.

emersion: Everything can stay except the roundtrip.

mortie:
Alright, I've changed the code to avoid extra roundtrips in the event handler. I also moved the `zxdg_output_manager_v1_get_xdg_output` to the `create_layer_surface` function, because that felt more in line with the existing style.

I also made the `surface->ready` state redundant, because the surface now initially loads the default image immediately, and then loads the correct image once the `zxdg_output_manager_v1_get_xdg_output` event fires.

If this looks good, I can squash my commits (I pushed prematurely a couple of times, sorry).

mortie:
Having looked at this more carefully, I don't think this solution is sound. Instead, it happens to work in my testing because the `zxdg_output_manager_v1_get_xdg_output` callback happen to fire before the `zwlr_layer_surface_v1_add_listener` callback fires, which isn't guaranteed AFAIK. As this patch stands now, `layer_surface_configure` could execute first, calling `render_frame_background` with the incorrect `surface->image` set, and then `handle_xdg_output_done` runs, setting the image to the correct image, but too late.

Some more logic is necessary to make sure that the `render_frame_background` and `render_frame` functions aren't called before A) both the zxdg callback and the layer_surface callback have fired if zxdg is supported, or B) only the layer_surface callback has fired if zxdg is not supported.

mortie:
https://github.com/swaywm/swaylock/pull/155/commits/7bcf95b21525bd67cc5b8024204fe39b7de39fe3 adds what I believe is the necessary ordering using an `events_pending` counter; adding an event listener increments it, an event listener firing decrements it, and only when it reaches 0 is the surface actually rendered.

mortie:
Having used this fix for a while, I think it breaks the forking behavior; it forks before the surfaces are ready. I'll work on a fix for that.

mortie:
After https://github.com/swaywm/swaylock/pull/155/commits/a0d398c8b8498cfadd550b5537b91952960c14b9, I'm not aware of any more issues, so I'm changing this PR to not be a draft again.

main.c
```diff
@@ -289,7 +312,15 @@ static void handle_xdg_output_description(void *data, struct zxdg_output_v1 *out
 }
 
 static void handle_xdg_output_done(void *data, struct zxdg_output_v1 *output) {
```
emersion, line 314: `xdg_output.done` is deprecated

mortie: https://github.com/swaywm/swaylock/pull/155/commits/853f292af0f6e1d7c675db8fff0ec2c49a05ffca moves the logic out of the xdg_output.done handler, and https://github.com/swaywm/swaylock/pull/155/commits/01a48ffff74eefc76dd15080df58b337f2d32e04 does synchronization using wl_display.sync instead of the ad-hoc events_pending counter.

main.c
```diff
@@ -118,11 +118,26 @@ static bool surface_is_opaque(struct swaylock_surface *surface) {
...
 
+struct zxdg_output_v1_listener _xdg_output_listener;
+static const struct wl_callback_listener surface_ready_listener;
+
 static void create_layer_surface(struct swaylock_surface *surface) {
 	struct swaylock_state *state = surface->state;
 
 	surface->image = select_image(state, surface);
 
+	static bool has_printed_zxdg_error = false;
```
emersion, line 129: It would be better to print this error in `main`. That way we wouldn't need this static variable.

main.c
```diff
@@ -1196,25 +1236,17 @@ int main(int argc, char **argv) {
...
-				"manager, images assigned to named outputs will not work");
-	}
-
 	struct swaylock_surface *surface;
 	wl_list_for_each(surface, &state.surfaces, link) {
 		create_layer_surface(surface);
 	}
 
+	wl_list_for_each(surface, &state.surfaces, link) {
+		while (surface->ready > 0) {
```
emersion, line 1245: This comparison with a bool looks weird. Maybe use `!surface->ready` instead?

main.c
```diff
@@ -1196,25 +1236,17 @@ int main(int argc, char **argv) {
...
+			wl_display_roundtrip(state.display);
```
emersion, line 1246: dispatch should be enough here, no need to roundtrip. Also please check for errors so that we don't enter in a busy-loop if the connection is broken (e.g. because of a protocol error or the compositor dying).

main.c
```diff
@@ -149,6 +164,14 @@ static void create_layer_surface(struct swaylock_surface *surface) {
 	zwlr_layer_surface_v1_add_listener(surface->layer_surface,
 			&layer_surface_listener, surface);
 
+	wl_surface_commit(surface->surface);
+
+	wl_callback_add_listener(
```
emersion, line 169: Could we add a comment explaining why we sync here? Something like: "We need to wait for the xdg_output.name event to be received before we can decide which image to display."

emersion:
Closing in favor of https://github.com/swaywm/swaylock/pull/330

PR #173, opened 2021-04-14
add support for custom PAM services

x1b6e6:


kennylevinsen:
Why would this be needed? Are you trying to run different swaylock instances using different stacks at the same time?

x1b6e6:
> Why would this be needed?

For add third-party pam services to swaylock such as fprintd or google totp without changing `swaylock` or built-in `login` pam service because it can be overwritten after updating.

> Are you trying to run different swaylock instances using different stacks at the same time?

Unfortunately I did not understand this question.

kennylevinsen:
> For add third-party pam services to swaylock such as fprintd or google totp without changing `swaylock` or built-in `login` pam service because it can be overwritten after updating.

It is a packaging error if /etc/pam.d/swaylock is overwritten by an upgrade. It is a system configuration file. Each file in /etc/pam.d/ is meant for the application by the same name. You're not supposed to piggyback on e.g. `login(1)`'s pam file.

kennylevinsen:
Closing - you're supposed to just edit `/etc/pam.d/swaylock` as appropriate, just like for every other auth service. File a bug with your package maintainer if the file is being overwritten.

PR #184, opened 2021-05-29
Draw ring and inner fill separately

### LINKED ISSUE #113: Inner Circle diameter not aligned
The inner circle's outer border aligns with the middle of the outer ring border. If choosing a transparent color for ring-color, then it looks a bit odd, with half of the inner circle shining though.
circle

loserMcloser:
Fixes #113.


emersion (approved):
Thanks!

PR #186, opened 2021-05-31
Fix indicator diameter calculation

loserMcloser:
arc_radius extends halfway into arc_thickness, so

`int buffer_diameter = (arc_radius + arc_thickness)  * 2;`

is incorrect (plus it doesn't take into account the outer indicator border).


loserMcloser:
Might want to add on an extra `+ 1` or `+ 2` onto the end of the formula for `buffer_diameter` as a safety buffer.

Issue #183, opened 2021-05-26
no ring while typing with sway 1.6-1 / libwlroots8 0.13.0-1


532910:
1. Under sway 1.5.1-2 / libwlroots7 0.12.0-2 swaylock works fine.
2. Under sway 1.6-1 / libwlroots8 0.13.0-1 there are no ring while typing.
3. `Cleared` ring is shown with both sway versions.
4. With `--indicator-radius=100` it works fine under sway 1.6-1.

Debian sid/experimental.

emersion:
Is this the same as https://github.com/swaywm/swaylock/issues/166? Are you seeing "Buffer size not divisible by scale" errors in the Sway debug log?

532910:
There no errors at all:
```
% swaylock -d -C /dev/null --color=000000                                                                                                                     
2021-05-27 11:52:11 - [main.c:1137] Found config at /dev/null
2021-05-27 11:52:11 - [main.c:1147] Parsing CLI Args
2021-05-27 11:52:11 - [main.c:280] output name is eDP-1
```
<details>
<summary> % WAYLAND_DEBUG=1 swaylock -d -C /dev/null --color=000000 </summary>

```
2021-05-27 11:56:09 - [main.c:1137] Found config at /dev/null
2021-05-27 11:56:09 - [main.c:1147] Parsing CLI Args
[3996335.647]  -> wl_display@1.get_registry(new id wl_registry@2)
[3996335.729]  -> wl_display@1.sync(new id wl_callback@3)
[3996335.991] wl_display@1.delete_id(3)
[3996336.055] wl_registry@2.global(1, "wl_shm", 1)
[3996336.110]  -> wl_registry@2.bind(1, "wl_shm", 1, new id [unknown]@4)
[3996336.179] wl_registry@2.global(2, "wl_drm", 2)
[3996336.226] wl_registry@2.global(3, "zwp_linux_dmabuf_v1", 3)
[3996336.271] wl_registry@2.global(4, "wl_compositor", 4)
[3996336.316]  -> wl_registry@2.bind(4, "wl_compositor", 3, new id [unknown]@5)
[3996336.380] wl_registry@2.global(5, "wl_subcompositor", 1)
[3996336.425]  -> wl_registry@2.bind(5, "wl_subcompositor", 1, new id [unknown]@6)
[3996336.488] wl_registry@2.global(6, "wl_data_device_manager", 3)
[...]

PR #187, opened 2021-05-31
Fix: Calculate buffer dimensions using scale factor

pinselimo:
The box_padding is adjusted for the display scaling but the actual dimensions of the text were not. This caused the indicator ring to disappear in some circumstances where the buffer size would not suffice to hold the data. Multiplying both the font height as well as the text width by surface->scale fixes these buffer dimensions.

This fixes at least my reproduction of #183 and also using swaylock-effects https://github.com/mortie/swaylock-effects/issues/53

### DIFF
--- a/render.c
+++ b/render.c
@@ -311,6 +311,10 @@ void render_frame(struct swaylock_surface *surface) {
 			}
 		}
 
+		// Ensure buffer size is multiple of buffer scale - required by protocol
+		new_height += surface->scale - (new_height % surface->scale);
+		new_width += surface->scale - (new_width % surface->scale);
+
 		if (buffer_width != new_width || buffer_height != new_height) {
 			destroy_buffer(surface->current_buffer);
 			surface->indicator_width = new_width;


emersion:
The rest of the code still uses `fe` and `extents` as if it was already multiplied by the scale. We could just multiply the value passed to `cairo_set_font_size` by the scale to fix everything I think (?), but it seems like Sway uses `pango_attr_scale_new` to set the scale instead. I don't remember what the implications are.

pinselimo:
It's weird indeed. The font itself was always scaled correctly. At least when it rendered.

After some tinkering I could trave the issue to the ````fe.height``` being an odd number. Specifically ```pool-buffer.c``` line 81 to 84 seems to be the issue.
I've provided a minimal fix to pin the problem down here: https://github.com/pinselimo/swaylock/commit/87cebb184f132b0d63212515c518781d20dbf7a0

> An odd value of fe.height leads to the indicator disappearing. It seems
that buffer_release is not called through the listener in this case.
As a result buffer->busy stays true.

> This commit provides the minimal changes required to resolve this
behavior. They all interact with the wayland shm buffer protocol.

Could this be a problem of the compositor? So far I had no success trying to fix this on the ```wlroots``` side and I am not too familiar with the wayland-server side.

emersion:
The Wayland protocol requires buffer sizes to be a multiple of the buffer scale.

pinselimo:
> The Wayland protocol requires buffer sizes to be a multiple of the buffer scale.

That's it, great! Thanks! 

(And sorry for my overexcited initial commit / PR - had been too happy to have fixed this annoying thing)

render.c
```diff
@@ -311,6 +311,10 @@ void render_frame(struct swaylock_surface *surface) {
 			}
 		}
 
+		// Ensure buffer size is multiple of buffer scale - required by protocol
+		new_height += surface->scale - (new_height % surface->scale);
+		new_width += surface->scale - (new_width % surface->scale);
```
emersion, line 316: If `new_width` is a multiple of `surface->scale`, this unnecessarily add `surface->scale` to the size. But this doesn't sound like a big deal, `new_width` isn't used for centering the sub-surface.

emersion (approved):
LGTM, thanks!

Issue #194, opened 2021-06-26
Ring not displayed


wb9688:
As of commit 05325b1197f9921f7d03096173589e5c06411ebb the ring is not being displayed when I start swaylock, turn off the only monitor (which is connected through DisplayPort), turn on that monitor and then try to enter my password. Logging in still works as expected, there just is no visual indicator.

gnull:
I see the same when I scale one of my monitor outputs. I.e. if I run Swaylock with

```
output eDP-1 scale 1.5
```

in my Sway config, the indicator ring does not show up. But if I set scale to 1, it works as expected.

minhduc0711:
@gnull fyi there is another issue on that https://github.com/swaywm/swaylock/issues/168.

PR #195, opened 2021-06-28
Fix indicator buffer not resizing after display powers off

mswiger:
I am having the same issue as described in https://github.com/swaywm/swaylock/issues/194 where the indicator ring would not show up after my displayed powered off. I did some debugging, and I noticed that after my display powers off, the value for `swaylock_state.auth_state` would never change from 0 (AUTH_STATE_IDLE). It is not clear to me why this is the case. However, one side-effect of this that I noticed is that the `indicator_width` and `indicator_height` state values will never change from their initial value of 0 in this situation, as the buffer resize logic in `render_frame` is unreachable.

By moving the buffer resize logic outside of the block that checks whether to render the indicator, the indicator buffers are able to be resized correctly, and everything in `swaylock` seems to resume correct behavior after my display is powered back on (the indicator ring renders, the auth state changes correctly, etc).

I am not sure if this is the correct fix, but if it is not, perhaps some of the information here will be helpful.

One other random observation: if I make the indicator ring show up before my display powers off by pressing a key, then power my display off and on again, the indicator ring will display correctly (even without this fix).

### DIFF
--- a/render.c
+++ b/render.c
@@ -321,6 +321,7 @@ void render_frame(struct swaylock_surface *surface) {
 		surface->indicator_width = new_width;
 		surface->indicator_height = new_height;
 		render_frame(surface);
+		return;
 	}
 
 	wl_surface_set_buffer_scale(surface->child, surface->scale);
--- a/render.c
+++ b/render.c
@@ -310,17 +310,17 @@ void render_frame(struct swaylock_surface *surface) {
 				new_width = extents.width + 2 * box_padding;
 			}
 		}
+	}
 
-		// Ensure buffer size is multiple of buffer scale - required by protocol
-		new_height += surface->scale - (new_height % surface->scale);
-		new_width += surface->scale - (new_width % surface->scale);
+	// Ensure buffer size is multiple of buffer scale - required by protocol
+	new_height += surface->scale - (new_height % surface->scale);
+	new_width += surface->scale - (new_width % surface->scale);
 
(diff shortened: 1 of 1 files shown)


render.c
```diff
@@ -310,13 +310,13 @@ void render_frame(struct swaylock_surface *surface) {
...
-			destroy_buffer(surface->current_buffer);
-			surface->indicator_width = new_width;
-			surface->indicator_height = new_height;
-			render_frame(surface);
-		}
+	if (buffer_width != new_width || buffer_height != new_height) {
+		destroy_buffer(surface->current_buffer);
+		surface->indicator_width = new_width;
+		surface->indicator_height = new_height;
+		render_frame(surface);
```
emersion, line 319: Can we add a `return` here to avoid attaching and committing the surface twice?

emersion (approved):
This sounds harmless enough to merge, even if I haven't dug up the exact cause of the size change. LGTM even without the comment fixed, because the original code already did a double-commit (feel free to submit a separate PR to add the `return` though!).

Thanks!

emersion:
Ah, needs a rebase.

render.c
mswiger: Done!

mswiger:
Thanks for taking a look! I have fixed the merge conflicts and added the return statement to prevent the double commit.

emersion:
Thanks for the fix!

Issue #181, opened 2021-05-20
Occassional unlock without password entered


savchenko:
This might be related to one of the [existing issues](https://github.com/swaywm/swaylock/issues?q=is%3Aissue+is%3Aopen+crash). From [#987360](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=987360):

> Sometimes when resuming from sleep, Swaylock will respond to the first keypress
of the password and display a spinner, but then freeze for about half a minute
and then just disappear and thereby allow access to Sway without the password
being entered.

Debian bug-report contains core-dumps.
Is there something in particular that can help with investigation of this bug?

kennylevinsen:
> Is there something in particular that can help with investigation of this bug?

The stack traces would be a start. Core dumps are not portable.

AdrianBunk:
Based on the provided coredumps it is:
```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x000055cc2e8068cd in loop_poll (loop=0x55cc30164780) at ../loop.c:111
#2  0x000055cc2e804eb5 in main (argc=2, argv=0x7ffebdfc1e78) at ../main.c:1235
(gdb) frame 1
#1  0x000055cc2e8068cd in loop_poll (loop=0x55cc30164780) at ../loop.c:111
111     in ../loop.c
(gdb) print *timer
$8 = {callback = 0x0, data = 0x55cc3015e010, expiry = {tv_sec = 59314,
    tv_nsec = 713617408}, link = {prev = 0x0, next = 0x0}}
(gdb) 
```
https://sources.debian.org/src/swaylock/1.5-2/loop.c/#L111

hjek:
I'm running Sway on Debian Unstable on two laptops, and the Swaylock crash seems to only happen on the older laptop.

savchenko:
Due to this bug `swaylock` is now removed from the upcoming Debian release: https://tracker.debian.org/news/1244697/swaylock-removed-from-testing/

hjek:
> Due to this bug `swaylock` is now removed from the upcoming Debian release

That is a pity. Sway is the best (and only) proper, lightweight Wayland compositor. Turns out [jwz was right back in '04](https://www.jwz.org/blog/2021/01/i-told-you-so-2021-edition/).

> If you are not running XScreenSaver on Linux, then it is safe to assume that your screen does not lock.

[...]

PR #209, opened 2021-09-19
Re-launch on crash

personalizedrefrigerator:
## Summary

 * Partially addresses #181.

## Caveat

This is an imperfect solution: After a crash, an attacker could briefly see and interact with the user's desktop while `swaylock` restarts.

## Requesting feedback on...

 * What should we do if `fork` fails when restarting `swaylock`? For example, if `ENOMEM`? (See the `TODO` in the source).
 * If we fork after connecting to the display, `state->display` seems to become invalid. Ideally, we would use the same display and keep input locked to `swaylock` between app restarts. Is there a way to do this?
 * Please also pay attention to: The code to handle restarts in the case where we're `daemonize`-ing.

## Testing it

 * I'm currently testing this PR by adding a `NULL`ptr deref to `main.c` that happens when authentication fails. 

<details><summary>It would be nice to have a script that tests this functionality.</summary>

 * Where would I put this?
 * In the script, it would probably be best to `sleep <timeout>; kill <child pid>`.
   * For this, a CLI option to print the child PID could be added to `swaylock`.

</details>


jirutka:
See also https://github.com/jirutka/swaylockd. Integrating re-launch mechanism into swaylock (not necessarily the same binary, but into the swaylock project) would be ofc a better solution.

I didn’t look into the source of this PR (yet), just cross-linking relevant resources.

emersion:
Superseded by https://github.com/swaywm/swaylock/pull/219

PR #210, opened 2021-09-27
Add documentation for input-clearing shortcuts

fuzxi:
Pressing Escape, Ctrl+C, or Ctrl+U clears all input, but this is not mentioned anywhere in the documentation.


emersion:
Hm, not sure I like the new "misc" section...

fuzxi:
I agree, but it also felt odd to double the size of "synopsis". Though I don't mind much either way as long as it's in there somewhere. LMK what you like best.

PR #216, opened 2021-11-19
Add editorconfig

ammgws:
Brought over from the sway repo. I only recently discovered editorconfig, but it has helped me finally get a grasp on the style issues I kept running into when submitting PRs.


emersion (approved):
Thanks

PR #219, opened 2021-12-20
Add support for ext-session-lock-v1 

emersion:
This is a new protocol to lock the session [1]. It should be more
reliable than layer-shell + input-inhibitor.

- [x] Bump wayland-protocols dep version

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/131

To build locally:

```sh
mkdir subprojects
git clone "git@gitlab.freedesktop.org:ifreund/wayland-protocols.git" -b ext-lock-v1 subprojects/wayland-protocols
meson configure build/ --force-fallback-for=wayland-protocols
```

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -89,6 +89,8 @@ struct swaylock_state {
 	int failed_attempts;
 	bool run_display;
 	struct zxdg_output_manager_v1 *zxdg_output_manager;
+	struct ext_session_lock_manager_v1 *ext_session_lock_manager_v1;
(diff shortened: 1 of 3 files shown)


ifreund (commented):
Works for me with the `ext-session-lock-v1` implementation in river here https://github.com/riverwm/river/pull/497 and the `wl_surface_commit()` call mentioned below removed when using `ext-session-lock-v1`.

emersion:
Addressed the comments, thanks for the review!

GreyXor:
I'll test

PR #225, opened 2022-02-10
seat: NULL-check on keyboard.modifiers event

tiosgz:
It can be sent before keyboard.keymap (for example
when running swaylock from a different tty), in
which case the xkb state isn't initialized yet.


emersion (approved):
Thanks!

Issue #212, opened 2021-10-14
[Feature request] Customizable text


rieje:
Simple feature: allow to customize text or +disable it+ (I suppose you can disable text by making it fully transparent). Can look aesthetically pleasing and is not really necessary  in most cases.

PR #230, opened 2022-03-08
Add possibility to update texts

Paulobergine:
This pull request allows users to update all the texts, when password is cleared, when caps lock is active, when password is verified and when password is wrong.


emersion:
What's the use-case for this?

Paulobergine:
> What's the use-case for this?

Just if you want to change the swaylock basic texts. In my case, I don't like the texts without maj : "clear", "wrong". And now, you can have "Clear", "Wrong", or others "key-words" as "Error", or every words.

emersion:
This doesn't strike me as a very useful feature.

emersion:
I'd be open to capitalize the text FWIW. The lowercase versions aren't a deliberate decision IIRC. Or maybe it's to mimic i3lock? What does i3lock do here?

For translations, I think something like https://github.com/swaywm/swaylock/pull/151 would be a better approach.

Paulobergine:
> I'd be open to capitalize the text FWIW. The lowercase versions aren't a deliberate decision IIRC. Or maybe it's to mimic i3lock? What does i3lock do here?
> 
> For translations, I think something like #151 would be a better approach.

Ok np, so for thoses like #212 , they may check this [repo](https://github.com/Paulobergine/swaylock-effects-text)

NiciTheNici:
I think this would be a useful feature for further customization. Is there any chance of this being merged anytime soon in the future?

Elu43:
It would be great for other languages. Is there any other way to change English ?
Thank you for your work

savente93:
Would also really appreciate this as I'd like to show the battery percentage on the lock screen so I can check it when my laptop is charging without having to unlock it every time.

PR #232, opened 2022-03-20
Delete whole utf8 character on backspace

johanmalm:
...rather than just the last byte of the password buffer.

Demostrate the need for this by taking the following steps:
- Apply the patch below (before this commit).
- Set keyboard layout to one where utf8 characters longer than one byte
  can be obtained, for example åäö using Swedish layout
  (XKB_DEFAULT_LAYOUT=se).
- Type "åäö" then press backspace and observe that it takes two
  backspaces to delete each character fully, and therefore six
  backspaces to fully clear the password buffer.

```
diff --git a/password.c b/password.c
index e1a1d9a..b640cd3 100644
--- a/password.c
+++ b/password.c
@@ -29,6 +29,7 @@ void clear_password_buffer(struct swaylock_password *pw) {
 static bool backspace(struct swaylock_password *pw) {
 	if (pw->len != 0) {
 		pw->buffer[--pw->len] = 0;
+		fprintf(stderr, "%s\n", pw->buffer);
 		return true;
 	}
 	return false;
```

### DIFF
--- a/include/unicode.h
+++ b/include/unicode.h
@@ -10,6 +10,14 @@
 #define UTF8_INVALID 0x80
 
 /**
+ * Gets the size in bytes of the last utf8 character in a NULL terminated string
+ * This function does not validate that the buffer contains correct utf8 data;
+ * it merely looks for the first byte that correctly denotes the beginning of a
+ * utf8 character.
+ */
+int utf8_last_size(const char *str);
+
+/**
  * Grabs the next UTF-8 character and advances the string pointer
  */
 uint32_t utf8_decode(const char **str);
--- a/password.c
+++ b/password.c
@@ -28,7 +28,8 @@ void clear_password_buffer(struct swaylock_password *pw) {
 
 static bool backspace(struct swaylock_password *pw) {
 	if (pw->len != 0) {
-		pw->buffer[--pw->len] = 0;
+		pw->len -= utf8_last_size(pw->buffer);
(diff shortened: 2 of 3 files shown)


include/unicode.h
```diff
@@ -9,6 +9,17 @@
...
+/**
+ * utf8_nr_bytes_to_prev_character - find distance to previous utf8 character
+ * @buf: C string buffer containing utf8 data
+ * @pos: position to start working backwards from
+ *
+ * This function does not validate that the buffer contains correct utf8 data;
+ * it merely looks for the first byte that correctly denotes the beginning of a
+ * utf8 character.
+ */
+int utf8_nr_bytes_to_prev_character(const char *buf, const char *pos);
```
emersion, line 21: Maybe `utf8_last_size` would be a better name, for consistency with `utf8_size`?

I don't think we need to pass two arguments, a single should be enough. `pw->buffer` is zero-terminated.

Nit: it's very nice to add docs, but we don't use `@`-style argument descriptions. Plain English sentences to describe args can be used instead.

johanmalm: Renamed to `utf8_last_size()`
Only pass `const char *str` and rely on it being `\0` terminated.
Local style used for documentation.
Have done `git commit --amend` and force-pushed. Hope that's okay.

emersion (approved):
LGTM, thanks!

PR #237, opened 2022-05-14
Unlock the screen on SIGUSR1

danieldg:
When the screen is locked using ext_session_lock, killing swaylock will
leave the session locked.  This is normally desirable if the kill is due
to the OOM killer or swaylock crashing, but can be undesirable if an
unlock is desired.  This adds a signal handler for SIGTERM if using
ext_session_lock without the --disable-unlock-on-sigterm option.

### DIFF
--- a/loop.c
+++ b/loop.c
@@ -82,7 +82,7 @@ void loop_poll(struct loop *loop) {
 	}
 
 	int ret = poll(loop->fds, loop->fd_length, ms);
-	if (ret < 0) {
+	if (ret < 0 && errno != EINTR) {
 		swaylock_log_errno(LOG_ERROR, "poll failed");
 		exit(1);
 	}
--- a/main.c
+++ b/main.c
@@ -5,6 +5,7 @@
 #include <fcntl.h>
 #include <getopt.h>
 #include <poll.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stdio.h>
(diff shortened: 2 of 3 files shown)


emersion:
Hm, I really don't think we should unlock the screen when the OOM killer kicks in.

An explicit Sway IPC command to unlock the screen would be less dangerous.

danieldg:
The OOM killer generally uses `SIGKILL`, so it won't trigger this.  However, there are other processes that might automatically send `SIGTERM`; changing the unlock signal to `SIGUSR1` would prevent accidental unlocking.

A sway IPC command would be sway-specific; swaylock is usable on other compositors.

danieldg:
> What is the use-case for the `--disable-unlock-on-sigusr` flag?

Disabling the feature if it is unwanted. I think this was more relevant when it was on SIGTERM - with SIGUSR1 it's not likely someone will be sending it signals by accident that you would want to disable.

emersion:
> Disabling the feature if it is unwanted. I think this was more relevant when it was on SIGTERM - with SIGUSR1 it's not likely someone will be sending it signals by accident that you would want to disable.

I'd prefer to leave that out then if there's no clear use-case -- every CLI flag we add, we can't remove easily without breaking users.

emersion (approved):
LGTM!

PR #240, opened 2022-05-30
Use mlock for password buffer

Molytho:
Only the memory buffer in the main process is locked. The buffer in the authentication child is not yet lock.

### DIFF
--- a/include/password-buffer.h
+++ b/include/password-buffer.h
@@ -1,0 +1,9 @@
+#ifndef _SWAY_PASSWORD_BUFFER_H
+#define _SWAY_PASSWORD_BUFFER_H
+
+#include <stddef.h>
+
+char *password_buffer_create(size_t size);
+void password_buffer_destroy(char *buffer, size_t size);
+
+#endif
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -67,7 +67,8 @@ struct swaylock_args {
 
 struct swaylock_password {
 	size_t len;
-	char buffer[1024];
+	size_t buffer_len;
+	char *buffer;
 };
 
 struct swaylock_state {
--- a/password-buffer.c
+++ b/password-buffer.c
@@ -1,0 +1,81 @@
+#define _POSIX_C_SOURCE 200809L
+#include "password-buffer.h"
+#include "log.h"
+#include "swaylock.h"
+#include <stdlib.h>
+#include <errno.h>
+#include <unistd.h>
+#include <limits.h>
+#include <sys/mman.h>
+
+static bool mlock_supported = true;
+static long int page_size = 0;
+
+static long int get_page_size() {
+	if (!page_size) {
+		page_size = sysconf(_SC_PAGESIZE);
+	}
+	return page_size;
+}
+
+// password_buffer_lock expects addr to be page alligned
+static bool password_buffer_lock(char *addr, size_t size) {
+	int retries = 5;
+	while (mlock(addr, size) != 0 && retries > 0) {
+		switch (errno) {
+		case EAGAIN:
+			retries--;
+			if (retries == 0) {
+				swaylock_log(LOG_ERROR, "mlock() supported but failed too often.");
+				return false;
+			}
+			break;
+		case EPERM:
+			swaylock_log_errno(LOG_ERROR, "Unable to mlock() password memory: Unsupported!");
+			mlock_supported = false;
+			return true;
+		default:
+			swaylock_log_errno(LOG_ERROR, "Unable to mlock() password memory.");
+			return false;
+		}
+		return false;
+	}
+
+	return true;
+}
+
+// password_buffer_unlock expects addr to be page alligned
+static bool password_buffer_unlock(char *addr, size_t size) {
+	if (mlock_supported) {
+		if (munlock(addr, size) != 0) {
+			swaylock_log_errno(LOG_ERROR, "Unable to munlock() password memory.");
+			return false;
+		}
+	}
+
+	return true;
+}
+
+char *password_buffer_create(size_t size) {
+	void *buffer;
+	int result = posix_memalign(&buffer, get_page_size(), size);
+	if (result) {
+		//posix_memalign doesn't set errno according to the man page
+		errno = result;
+		swaylock_log_errno(LOG_ERROR, "failed to alloc password buffer");
+		return NULL;
+	}
+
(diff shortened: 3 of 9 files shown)


password-buffer.c
```diff
@@ -0,0 +1,50 @@
...
+
+void password_buffer_destroy(char* buffer, size_t size) {
+	clear_buffer(buffer, size);
+	password_buffer_unlock(buffer, size);
+	free(buffer);
+}
+
+bool password_buffer_lock(char* buffer, size_t size) {
+	#ifdef __linux__
+	// Most non-linux platforms require root to mlock()
```
emersion, line 33: I don't think that's the case. The amount of memory a process can `mlock` is in general controlled by `RLIMIT_MEMLOCK`.

include/password-buffer.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _SWAY_PASSWORD_BUFFER_H
+#define _SWAY_PASSWORD_BUFFER_H
+
+#include <stddef.h>
+#include <stdbool.h>
+
+char* password_buffer_create(size_t size);
+void password_buffer_destroy(char* buffer, size_t size);
```
emersion, line 8: Style: `*` on the right (applies to the rest of the patch)

password-buffer.c
kennylevinsen: Indeed. On Linux you can `mlock` up to the configurable RLIMIT_MEMLOCK without privileges, while more than that requires `CAP_IPC_LOCK`. The restrictions are similar on FreeBSD, except if `security.bsd.unprivileged_mlock` is set to 0.

Molytho: I just copy-pasted this. I will delete the ifdef and add some error handling instead

include/password-buffer.h
```diff
@@ -0,0 +1,12 @@
...
+bool password_buffer_lock(char* buffer, size_t size);
+bool password_buffer_unlock(char* buffer, size_t size);
```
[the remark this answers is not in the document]
Molytho: How should we handle the buffer in struct swaylock_password? Currently it is mlocked manually. I wanted to factor this out to use a common function for both existing password buffers.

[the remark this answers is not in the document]
emersion: We should be able to change `char buffer[1024]` in `struct swaylock_password` to be a `char *`, and then use `password_buffer_create`.

include/password-buffer.h
```diff
@@ -0,0 +1,12 @@
...
+
+#include <stddef.h>
+#include <stdbool.h>
+
+char *password_buffer_create(size_t size);
+void password_buffer_destroy(char *buffer, size_t size);
+bool password_buffer_lock(char *buffer, size_t size);
+bool password_buffer_unlock(char *buffer, size_t size);
+
+#endif
```
emersion, line 12: Style: missing final newline

password-buffer.c
```diff
@@ -0,0 +1,85 @@
...
+		void *addr = calculate_page_boundary(buffer);
+		size_t size = ((size_t)buffer + buffer_size) - (size_t)addr;
+		if (munlock(addr, size) != 0) {
+			swaylock_log_errno(LOG_ERROR, "Unable to munlock() password memory.");
+			return false;
+		}
+	}
+
+	return true;
+}
```
emersion, line 85: Ditto

password-buffer.c
```diff
@@ -0,0 +1,85 @@
...
+			return false;
+		}
+		return false;
+	}
+
+	return true;
+}
+
+bool password_buffer_unlock(char *buffer, size_t buffer_size)
+{
```
emersion, line 74: Style: `{` should be on the previous line

include/password-buffer.h
```diff
@@ -0,0 +1,12 @@
+#ifndef _SWAY_PASSWORD_BUFFER_H
+#define _SWAY_PASSWORD_BUFFER_H
+
+#include <stddef.h>
+#include <stdbool.h>
+
+char *password_buffer_create(size_t size);
+void password_buffer_destroy(char *buffer, size_t size);
+bool password_buffer_lock(char *buffer, size_t size);
+bool password_buffer_unlock(char *buffer, size_t size);
```
emersion, line 10: lock/unlock can be made private & static now.

password-buffer.c
```diff
@@ -0,0 +1,79 @@
...
+}
+
+void password_buffer_destroy(char *buffer, size_t size) {
+	clear_buffer(buffer, size);
+	password_buffer_unlock(buffer, size);
+	free(buffer);
+}
+
+static void *calculate_page_boundary(void *addr) {
+	long int PAGE_SIZE = sysconf(_SC_PAGESIZE);
```
emersion, line 34: Use lower-case since it's not a global constant?

password-buffer.c
```diff
@@ -0,0 +1,79 @@
...
+	size_t page_offset = ((size_t)addr) % PAGE_SIZE;
+	return addr - page_offset;
+}
+
+bool password_buffer_lock(char *buffer, size_t buffer_size) {
+	int retries = 5;
+	void *addr = calculate_page_boundary(buffer);
+	size_t size = ((size_t)buffer + buffer_size) - (size_t)addr;
```
emersion, line 42: Nit: the correct type to use here is `uintptr_t`.

emersion:
One last thing: can we use `posix_memalign` instead of `malloc`? This should save us from the offset computation.

include/swaylock.h
```diff
@@ -1,5 +1,8 @@
 #ifndef _SWAYLOCK_H
 #define _SWAYLOCK_H
+
+#define _POSIX_C_SOURCE 200809L
```
emersion, line 4: We shouldn't define `_POSIX_C_SOURCE` in a header. Only in a C source file.

password-buffer.c
```diff
@@ -0,0 +1,80 @@
...
+#include <errno.h>
+#include <unistd.h>
+#include <limits.h>
+#include <sys/mman.h>
+
+
+static bool mlock_supported = true;
+static long int page_size;
+
+static void __attribute__((constructor)) password_buffer_init() {
```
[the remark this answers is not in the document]
Molytho: Sorry. Thought this works with other compilers too.

emersion (approved):
LGTM, thanks for bearing with me!

Issue #241, opened 2022-05-31
Accept input for the next password attempt while checking the password


ddevault:
When I enter the wrong password, it would be nice to be able to type another attempt while it's being validated, so I don't have to wait for it to tell me that it's wrong.

PR #252, opened 2022-09-02
Accept input while validating

WhyNotHugo:
Allow typing new input while the previous one is validating. Can be
tested with:

    ninja -C build && sway -c sway.conf

Where `sway.conf` is:

    exec ./build/swaylock

Fixes: https://github.com/swaywm/swaylock/issues/241

### DIFF
--- a/password.c
+++ b/password.c
@@ -95,10 +95,6 @@ static void submit_password(struct swaylock_state *state) {
 
 void swaylock_handle_key(struct swaylock_state *state,
 		xkb_keysym_t keysym, uint32_t codepoint) {
-	// Ignore input events if validating
-	if (state->auth_state == AUTH_STATE_VALIDATING) {
-		return;
-	}
 
 	switch (keysym) {
 	case XKB_KEY_KP_Enter: /* fallthrough */


password.c
```diff
@@ -78,28 +78,40 @@ static void schedule_password_clear(struct swaylock_state *state) {
 			state->eventloop, 10000, clear_password, state);
 }
 
+static struct swaylock_password *copy_password(struct swaylock_password pw) {
+	struct swaylock_password *copy = malloc(sizeof(pw));
+
+	copy->len = pw.len;
+	copy->buffer_len = pw.buffer_len;
+	copy->buffer = malloc(pw.len + 1);
```
emersion, line 86: This is insecure. Use `password_buffer_create` instead.

password.c
```diff
@@ -78,28 +78,40 @@ static void schedule_password_clear(struct swaylock_state *state) {
...
 
 	state->auth_state = AUTH_STATE_VALIDATING;
 
-	if (!write_comm_request(&state->password)) {
+	// Create a copy to pass for validation.
+	struct swaylock_password *pw = copy_password(state->password);
+	// And free the buffer for anything that's typed during validation or later.
+	clear_password_buffer(&state->password);
+
+	if (!write_comm_request(pw)) {
```
emersion, line 104: The write is non-blocking. Why do we need to copy?

WhyNotHugo: My bad, I was under the impression that we might alter the buffer while before it's written out, but this does actually block very briefly.

The updated approach works, but I'm noticing it allows trying a lot of attempts without any delay before write_comm_request between them. Even if pam complains `pam_authenticate failed: maximum number of authentication tries exceeded`, writing the right password without any pause after many failed tries still works. I think this would allow using a fake USB keyboard that just brute forces without limitation.

emersion:
Alright, let's give this a try. Thanks for having a look @mstoeckl!

PR #253, opened 2022-09-10
Fix loss of clarity with odd-sized centered image

jlindgren90:
When using `scaling=center`, an image with an odd width or height is currently shifted by 0.5 pixel relative to display pixels, resulting in a loss of detail when rendered. This is particular noticeable with an image containing text, which becomes noticeably blurry.

It's true that the current code is "accurate" in that it exactly centers the image on the display, but I think most users would probably prefer to live with the image being off-center by 0.5 pixel in return for pixel-perfect rendering.

### DIFF
--- a/background-image.c
+++ b/background-image.c
@@ -100,9 +100,14 @@ void render_background_image(cairo_t *cairo, cairo_surface_t *image,
 		break;
 	}
 	case BACKGROUND_MODE_CENTER:
+		/*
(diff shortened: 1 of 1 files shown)


jlindgren90:
`round()` works too; I wasn't sure if there would be concerns about the `-lm` dependency.

emersion (approved):
Ah, right, forgot that we don't link to libm yet. Well, probably not worth the effort.

Thanks!

PR #255, opened 2022-09-24
Use single pixel buffer and viewporter for solid backgrounds

Nefsen402:


emersion:
The first two commits LGTM, pushed to master.

PR #256, opened 2022-11-06
Fix capitalization.

digitalwtf:


NiciTheNici:
This seems like an arbitrary change based on personal preference. #230 would address this issue aswell.

emersion (approved):
Thanks!

PR #257, opened 2022-11-12
Ensure that ext_session_lock_v1.unlock_and_destroy is processed.

alebastr:
Use wl_display_roundtrip to ensure that the unlock request is received and processed by the server. The protocol requires[1] this to avoid possible races.

[1]: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/161

Without the patch I regularly observed the red screen; WAYLAND_DEBUG output had no messages after unlock_and_destroy, so I assumed this was the cause. A month of testing with no unlock issues seems to be a good confirmation.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1285,7 +1285,7 @@ int main(int argc, char **argv) {
 
 	if (state.ext_session_lock_v1) {
 		ext_session_lock_v1_unlock_and_destroy(state.ext_session_lock_v1);
-		wl_display_flush(state.display);
+		wl_display_roundtrip(state.display);
 	}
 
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

jpalus:
Upgraded sway yesterday from `1.7` to `1.8-rc1` and today experienced "red screen" for the very first time. I suppose there is a possibility this PR might be fixing it, so perhaps it'd be good idea to release `1.6.1` now that `sway 1.8` is around the corner?

emersion:
Yes.

PR #261, opened 2022-11-17
Initialize pw backend after parsing CLI args

raboof:
To make sure the pw backend picks up the loglevel when passing `--debug`.

(draft because swaylock isn't working for me yet - I wanted debug logging for a reason ;) )


emersion:
This has security implications: it executes more code as root. In particular, with this patch, files are opened as root.

raboof:
ah, using pam here, indeed I didn't notice this is the case for the shadow backend :thinking:

raboof:
OK, so:
* we do want a way to enable debug logging of the backend (right?)
* we don't want to open files before dropping privileges
* we don't want to change the logic of first processing the config file and then processing the CLI parameters
* we don't want to introduce unnecessary changes in the way the pam and shadow backends need to be invoked
* we don't want to introduce other mechanisms (such as setting environment variables) to enable debugging, as that'd be inconsistent with related tools like sway itself

Perhaps we could have the forked process look for the --debug flag itself, ignoring any other flags and the config file? Somewhat inconsistent and not very DRY, but possibly acceptable since '--debug' is a somewhat 'special' diagnostic flag anyway? WDYT?

emersion:
Yeah that sounds reasonable!

emersion:
Hm, unfortunately this is not quite enough. `swayidle -- --debug` for instance processes `--debug` as an argument, not an option.

Can we use `getopt` to parse `--debug`?

raboof:
> Can we use `getopt` to parse `--debug`?

updated

emersion:
We need to reset `optind = 1` to allow `getopt` to be called multiple times.

emersion:
Hm, we need to reset `optind = 1` between the two `getopt` loops, ie. after the `log_init` call.

raboof:
> We need to reset `optind = 1` to allow `getopt` to be called multiple times.

Gotcha! Can confirm this indeed makes cases where `--debug` isn't the first parameter work :laughing: . Apologies for not testing sufficiently...

raboof:
> Hm, we need to reset `optind = 1` between the two `getopt` loops, ie. after the `log_init` call.

That was already the case (`parse_options` set `optind = 1` before entering the loop that calls `getopt_long`). TBH this is pretty surprising API to me, but also setting `optind = 1` before entering the loop in `log_init` as well did improve behavior :thinking:

emersion (approved):
Oh right, indeed! LGTM.

You're correct that the `getopt` API is really not great, but oh well…

raboof:
Thanks for the thorough review!

main.c
```diff
@@ -1135,8 +1135,34 @@ static void term_in(int fd, short mask, void *data) {
...
+void log_init(int argc, char **argv) {
+	static struct option long_options[] = {
+		{"debug", no_argument, NULL, 'd'},
+        {0, 0, 0, 0}
+    };
+    int c;
+	optind = 1;
+    while (1) {
+		int opt_idx = 0;
+		c = getopt_long(argc, argv, "d", long_options, &opt_idx);
```
lilyinstarlight, line 1151: It looks like this PR broke option parsing for swaylock on my system

```suggestion
		c = getopt_long(argc, argv, ":d", long_options, &opt_idx);
```

Without the above change, swaylock still does not properly parse arguments unless `--debug` is included in the argument list somewhere, and even then it still prints a confusing error if `--debug` is not the very first argument

With the above change it does not print an error if `--debug` is not included and it does properly parses the other arguments again

From the getopt manual page:

> *  If  the first character (following any optional '+' or '-' described above) of optstring is a colon (':'), then getopt() likewise does not print an error message.  In addition, it returns ':' instead of '?' to indicate a missing option argument.  This allows the caller to distinguish the two different types of errors.

raboof: Ugh, I'm so sorry about that. I swear I tested a number of variations, but obviously not sufficiently. I'll get on testing if adding `:` indeed seems sufficient when I try a number of different invocation variations, and then PR.

lilyinstarlight: No worries! I'm actually starting to notice still odd behavior even with this change, so doing `getopt_long` twice seems more of a beast than I thought

lilyinstarlight: Specifically I'm still seeing stuff like this, but only if `--debug` is not the first arg 🤔 (same behavior when it is excluded too):

```
[I] lily@bina ~/s/swaylock (fix/no-debug)> build/swaylock -c 000000 --bs-hl-color 333333 --debug
2023-01-25 11:19:53 - [main.c:1214] Parsing CLI Args
2023-01-25 11:19:53 - [main.c:39] Invalid color --bs-hl-color, defaulting to 0xFFFFFFFF
```

lilyinstarlight: > **By default, getopt() permutes the contents of argv as it scans**, so that eventually all the nonoptions are at the end.  Two other scanning modes are  also  implemented.  If the first character of optstring is '+' or the environment variable POSIXLY_CORRECT is set, then option processing stops as soon as a nonoption argument is encountered.  If '+' is not the first character of optstring, it is treated as a normal option.  If POSIXLY_CORRECT behaviour is required in this case optstring will contain two '+' symbols.  If the first character of optstring is '-', then each nonoption argv-element is handled as if it were the argument of an option with character code 1.

Ugh, I guess we need `-` in front of it too maybe

raboof: ok, so it seems the first time around (in `log_init`) it actually mutates `argv` because there it doesn't know `-c` should take an argument. That seems like borderline a bug in getopt tbh... but I guess we should work around it in swaylock anyway.

raboof: oh you beat me to it, I should refresh more often :laughing:

lilyinstarlight: Sweet! I opened #279

PR #275, opened 2023-01-12
Print PID of child process after forking

WhyNotHugo:
This lets a parent track the lifetime of a child when using `--daemonize`. Using `--daemonize` is the only way for a parent to determine when the screen has been locked.

### DIFF
--- a/main.c
+++ b/main.c
@@ -61,11 +61,12 @@ int lenient_strcmp(char *a, char *b) {
 
 static void daemonize(void) {
 	int fds[2];
+	int pid;
 	if (pipe(fds) != 0) {
 		swaylock_log(LOG_ERROR, "Failed to pipe");
 		exit(1);
 	}
-	if (fork() == 0) {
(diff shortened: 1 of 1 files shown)


kennylevinsen:
This is racy, and a parent script can background non-daemonized swaylock and look at `jobs` to see its PID.

WhyNotHugo:
If I run swaylock non-daemonised, the parent has no way of knowing when the session is actually locked.

How exactly is it racey? Do you mean due to potential PID reuse?

emersion:
Related: #42

WhyNotHugo:
#42 does seem like a better solution, though I fear it might be stuck in discussion for a long time still.

I think I can use `prctl` with `PR_SET_CHILD_SUBREAPER` to track the lifetime of the child process in the meantime. If that works (at least for now), I'll go ahead and close this.

emersion:
Related: #281

WhyNotHugo:
Closing in favour of #281

PR #277, opened 2023-01-14
Refactor render_frame

mstoeckl:
This PR reorganizes the `render_frame` function so that it no longer calls itself recursively. (This sometimes made swaybg render the same indicator twice.) To estimate font and text extents, it now uses a "test" cairo context based on a 1x1 surface.

The following additional patch can be used to make the buffer size visible:
```
diff --git a/render.c b/render.c
index 3ce7a80..3e5e6eb 100644
--- a/render.c
+++ b/render.c
@@ -211,7 +211,7 @@ void render_frame(struct swaylock_surface *surface) {
 
        // Clear
        cairo_save(cairo);
-       cairo_set_source_rgba(cairo, 0, 0, 0, 0);
+       cairo_set_source_rgba(cairo, 0, 0, 0, 0.5);
        cairo_set_operator(cairo, CAIRO_OPERATOR_SOURCE);
        cairo_paint(cairo);
        cairo_restore(cairo);
```

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -86,6 +86,8 @@ struct swaylock_state {
 	struct swaylock_args args;
 	struct swaylock_password password;
 	struct swaylock_xkb xkb;
+	cairo_surface_t *test_surface;
+	cairo_t *test_cairo; // used to estimate font/text sizes
 	enum auth_state auth_state;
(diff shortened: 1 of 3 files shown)


emersion:
Nice cleanup, thanks!

PR #279, opened 2023-01-25
Fix option parsing without --debug after #261

lilyinstarlight:
Option parsing was messed up by #261 because `getopt_long` mutates `argv` by default, and it printed confusing error messages when `--debug` was not the first option. This PR fixes both of those issues

See https://github.com/swaywm/swaylock/pull/261#discussion_r1086819313

From getopt manual page:

> By default, getopt() permutes the contents of argv as it scans, so that eventually all the nonoptions are at the end. Two other scanning modes are also implemented. [...] If the first character of optstring is '-', then each nonoption argv-element is handled as if it were the argument of an option with character code 1.

> If the first character (following any optional '+' or '-' described above) of optstring is a colon (':'), then getopt() likewise does not print an error message. In addition, it returns ':' instead of '?' to indicate a missing option argument. This allows the caller to distinguish the two different types of errors.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1148,7 +1148,7 @@ void log_init(int argc, char **argv) {
 	optind = 1;
     while (1) {
 		int opt_idx = 0;
-		c = getopt_long(argc, argv, "d", long_options, &opt_idx);
+		c = getopt_long(argc, argv, "-:d", long_options, &opt_idx);
 		if (c == -1) {
 			break;
 		}


raboof (approved):
sorry again for the breakage, this indeed both logically makes sense and works for my test cases as well :+1:

emersion:
Hm, unfortunately the leading `-` is a GNU extension, and we try to not use these. Does it work to extract the `optstring` into a constant used by both `getopt` calls?

lilyinstarlight:
> Hm, unfortunately the leading `-` is a GNU extension, and we try to not use these. Does it work to extract the `optstring` into a constant used by both `getopt` calls?

Oh yeah I meant to address that because I did consider whether or not to use a GNU extension, but it turns out `getopt_long` itself is also a GNU extension

So I'll check later today and see if this will cause a regression with like musl or something

lilyinstarlight:
> Hm, that's fair. musl seems to handle `optstring[0] == '-'` correctly.

For what it's worth, it looks like FreeBSD libc, NetBSD libc, newlib, and even uclibc also implement these GNU extensions correctly. The only one that looked like it didn't implement it that I looked at was dietlibc

So I'd say it's a pretty safe GNU extension to use, but I'm also happy to try factoring out the optstring if you would rather

Thank you @emersion!

emersion (approved):
Thanks!

PR #266, opened 2022-11-27
Drop support for layer-shell

emersion:
Superseded by ext-session-lock-v1

PR #280, opened 2023-01-27
Wait for the locked event

emersion:
We were ignoring the locked event. Wait for it instead.

### DIFF
--- a/main.c
+++ b/main.c
@@ -313,7 +313,8 @@ struct wl_output_listener _wl_output_listener = {
 };
 
 static void ext_session_lock_v1_handle_locked(void *data, struct ext_session_lock_v1 *lock) {
-	// Who cares
+	struct swaylock_state *state = data;
+	state->locked = true;
 }
 
 static void ext_session_lock_v1_handle_finished(void *data, struct ext_session_lock_v1 *lock) {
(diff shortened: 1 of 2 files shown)


main.c
```diff
@@ -1295,8 +1296,22 @@ int main(int argc, char **argv) {
...
+				swaylock_log(LOG_ERROR, "wl_display_dispatch() failed");
+				return 2;
+			}
+		}
+	} else {
+		if (wl_display_roundtrip(state.display) < 0) {
+			swaylock_log(LOG_ERROR, "wl_display_roundtrip() failed");
+			return 2;
+		}
+		state.locked = true;
```
WhyNotHugo, line 1311: I don't think this statement is correct; this should only be set when the locked event has been received from the compositor (as done in `ext_session_lock_v1_handle_locked`).

emersion: This is the legacy code-path, when ext-session-lock-v1 is unused (ie, when layer-shell is used). I've only included it so that this patch can be included in a bugfix release, we'll soon drop that logic: #266.

WhyNotHugo (approved):
Works for me (with `ext_session_lock_manager_v1`).

main.c
kennylevinsen, line 1311: If we set this in the layer shell configure handler, we save branching here and get a better indication of locking than a single blind roundtrip.

On the other hand, this code will be deleted almost immediately so I don't care too much about it... :)

main.c
```diff
@@ -1295,8 +1296,22 @@ int main(int argc, char **argv) {
 		create_surface(surface);
 	}
 
+	if (state.ext_session_lock_manager_v1) {
+		while (!state.locked) {
```
kennylevinsen, line 1300: Hmm, if we do not enter the locked state for whatever reason, this turns into an infinite loop. I wonder if we should have a timeout here...

kennylevinsen (approved):
LGTM with optional comments

main.c
emersion: We get a `finished` event in that case, and exit.

main.c
emersion: I'd rather keep the changes for the legacy code-path minimal.

main.c
kennylevinsen: Yeah, fair enough, we already set up the lock and no need to code around a broken compositor.

emersion:
Thanks!

PR #285, opened 2023-02-07
Option to use bold font weight

### LINKED ISSUE #182: unable to set bold font
1. `font Fantasque Sans Mono Bold 9` in `.config/sway/config` works fine
2. `swaynag -f 'Fantasque Sans Mono Bold 20'` works fine
3. `swaylock --font "Fantasque Sans Mono" works fine
4. `swaylock --font "Fantasque Sans Mono Bold" uses a default font

metafates:
This PR adds a boolean argument `--font-bold`. If it is present, swaylock will use `CAIRO_FONT_WEIGHT_BOLD` as specified [here](https://www.cairographics.org/manual/cairo-text.html#cairo-font-weight-t)

Closes #182


hramrach:
It would be nice to support all font variants, rather than a hardcoded subset.

How do other application support those fonts that come in half dozen different weights and 3 different variants?

hramrach:
From the docs:
> Note: The cairo_select_font_face() function call is part of what the cairo designers call the "toy" text API. It is convenient for short demos and simple programs, but it is not expected to be adequate for serious text-using applications.

> For "real" font selection, see the font-backend-specific font_face_create functions for the font backend you are using. (For example, if you are using the freetype-based cairo-ft font backend, see cairo_ft_font_face_create_for_ft_face() or cairo_ft_font_face_create_for_pattern().) The resulting font face could then be used with cairo_scaled_font_create() and cairo_set_scaled_font().

>Similarly, when using the "real" font support, you can call directly into the underlying font system, (such as fontconfig or freetype), for operations such as listing available fonts, etc.

>It is expected that most applications will need to use a more comprehensive font handling and text layout library, (for example, pango), in conjunction with cairo.

hramrach:
This is likely the expected interface https://gtk.developpez.com/doc/en/pango/pango-Fonts.html#pango-font-description-from-string

It can parse the different font variants and options from free-form text description, there is no need to have a separate program parameter for each.

PR #287, opened 2023-02-20
Choose random file if pointed to a directory for image

Jeinzi:
Hey!

With this commit, if --image gets pointed to a directory, swaylock chooses a random file from it.
I'd be glad if this could be merged :)

Best regards,
Julian

### DIFF
--- a/main.c
+++ b/main.c
@@ -12,6 +12,8 @@
 #include <string.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <dirent.h>
+#include <limits.h>
 #include <time.h>
 #include <unistd.h>
 #include <wayland-client.h>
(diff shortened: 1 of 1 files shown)


emersion:
Sorry, but I'd rather not add this feature. This is better done by a script which does the randomization and then invokes swaylock.

Jeinzi:
While I agree that a script is a good way to do this, I want to emphasize that swaylock currently just ignores the parameter and does something else in case a directory is supplied. This feature does something sensible in this situation, in an intuitive manner I would say, and I image choosing one image at random from a collection is a common use case.
Anyway, I'll try my luck over at swaylock-effects :)

emersion:
The expected behavior is to error out. If swaylock doesn't do that, please open a bug report.

Jeinzi:
Well, it prints an error message and falls back to a white background.

Issue #180, opened 2021-05-18
Swaylock can run out of buffers when output is resized


mstoeckl:
When running Sway nested under wayland or x11, it is possible to dynamically resize the outputs. Most of the time, `swaylock` correctly redraws when this is done. However, if rendering a frame takes a long time -- which can happen when running `swaylock -i large_image.png` -- sometimes `swaylock` fails to provide a properly sized buffer for the new output size. In this case, the background will only cover a portion of the screen, revealing whatever is underneath the rest.

This behavior stems from the limited number of background buffers associated with a given `struct swaylock_surface`. If swaylock is made to render a new frame background at least three times in short succession, it will render and submit two buffers, but will run out of buffers on the third call to [`render_frame_background`](https://github.com/swaywm/swaylock/blob/master/render.c#L35-L48). This can happen if the output is resized many times in a second, so that swaylock receives a large number of layer surface configure events in one batch; in that case, it will try to render a new frame background for each.

Issue #282, opened 2023-01-29
Red screen & unable to unlock when changing screen


jvanbruegge:
This is with ext-session-lock (swaylock 1.7) and sway 1.8. If I lock my laptop, suspend it by closing it, then connect it to my docking station with two external monitors and resuming it, I get greeted by completely red screens on both monitors and no way to unlock except rebooting the machine.

I can see this in the logs:
```
[main.c:1188] Parsing CLI Args
[main.c:489] Loaded image <path> for output *
[main.c:1245] Using ext-session-lock-v1

<other logs>

ext_session_lock_surface_v1@33: error 2: committed surface dimensions do not match last acked configure
```

Not sure if that is related, but shortly after, waybar dumps core.

kennylevinsen:
This means that swaylock hit a protocol error, stemming from this check: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_session_lock_v1.c#L145

If you can reproduce with swaylock 1.7.2 (or master), run swaylock manually with `WAYLAND_DEBUG=1` set in the environment, collecting output to a file as it crashes. This should show what causes the protocol error..

jacksgt:
I am facing the exact same issue and started debugging, but then I realized that I'm in fact using `swaylock-effects` (1.6.10) instead of the regular `swaylock`.
With vanilla `swaylock` (1.7) screen locking works without issues on `sway` 1.8.

I see that there is another discussion for swaylock-effects: https://github.com/mortie/swaylock-effects/issues/97

jvanbruegge:
Yes, it is really hard for me to reproduce and I am not even sure if 1.7.2 might have fixed it, can't remember if the last red screen was before or after the upgrade

jvanbruegge:
Ok, I just had this today again, so it is indeed still an issue. Sadly I had no time for investigation so I have to wait until the next time this happens

fberg:
This happens fairly frequently to me, and I captured logs with `WAYLAND_DEBUG=1`. Happened under similar circumstances as in the OP (suspend with internal display, wakeup with external monitors attached). I'm on swaylock 1.7.2.

Btw: you can unlock the session by switching to a different TTY and starting swaylock there (with appropriate `WAYLAND_DISPLAY` set).

[swaylock.error.log](https://github.com/swaywm/swaylock/files/12567891/swaylock.error.log)

kennylevinsen:
[...]

PR #288, opened 2023-03-12
Stop pooling background surface buffers

mstoeckl:
Quoting the commit message:
> The wl_buffers for the background surface only need to be updated when the output dimensions change. Using the fixed pool of two buffers to cache these buffers does not help, since if a new buffer is needed, it will have a different size than whatever buffers were cached. Furthermore, because the pool has fixed size, it is possible to run out of buffers if configure events arrive faster than pool buffers  are marked not busy,, which can lead to protocol errors when the background surface is committed after acknowledging a new size, but without attaching a buffer that matches that size.

Without this change, one way to make Swaylock run out of buffers is to run it inside a nested instance of Sway, configured to draw a large image as the background, and then rapidly resize the nested instance until Swaylock receives several configure events in a single `wl_display_dispatch()` call.

This fixes https://github.com/swaywm/swaylock/issues/180 and _might_ help with https://github.com/swaywm/swaylock/issues/282.

The new buffer allocation strategy for the background is closer to what swaybg does.

### DIFF
--- a/render.c
+++ b/render.c
@@ -4,6 +4,7 @@
 #include "cairo.h"
 #include "background-image.h"
 #include "swaylock.h"
+#include "log.h"
 
 #define M_PI 3.14159265358979323846
 const float TYPE_INDICATOR_RANGE = M_PI / 3.0f;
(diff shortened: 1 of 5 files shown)


render.c
```diff
@@ -41,30 +42,40 @@ void render_frame_background(struct swaylock_surface *surface) {
...
 
-	struct pool_buffer *buffer = get_next_buffer(state->shm,
-			surface->buffers, buffer_width, buffer_height);
-	if (buffer == NULL) {
-		return;
-	}
+	if (buffer_width != surface->last_buffer_width ||
+		buffer_height != surface->last_buffer_height) {
+		surface->last_buffer_width = buffer_width;
+		surface->last_buffer_height = buffer_height;
```
emersion, line 48: We should probably do this after the `create_buffer()` call succeeds?

render.c
```diff
@@ -41,30 +42,40 @@ void render_frame_background(struct swaylock_surface *surface) {
 		return; // not yet configured
 	}
 
-	struct pool_buffer *buffer = get_next_buffer(state->shm,
-			surface->buffers, buffer_width, buffer_height);
-	if (buffer == NULL) {
-		return;
-	}
+	if (buffer_width != surface->last_buffer_width ||
+		buffer_height != surface->last_buffer_height) {
```
emersion, line 46: Style nit: two tabs for `if `continuation lines, to be able to tell apart the condition and the body

emersion:
Looks good to me apart from this

emersion (approved):
Thanks!

PR #292, opened 2023-04-02
Synchronize highlight position between outputs

mstoeckl:
This change makes Swaylock look slightly nicer, by ensuring that highlight positions are updated once per key stroke, and not once per time a frame is drawn -- which could make highlight positions change differently depending on the number of active monitors.

For example, with two outputs, _before_ this change, the highlight position would change independently for each output, and often by small amounts due to it being updated twice / rotated by two approximate half turns:
[bad.webm](https://user-images.githubusercontent.com/7674289/229323465-13a21727-e985-4bdc-a820-de43bad7402b.webm)
And _after_ this change:
[good.webm](https://user-images.githubusercontent.com/7674289/229323476-8a80889d-478c-4d2b-a077-933ac9a713d8.webm)

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -87,6 +87,7 @@ struct swaylock_state {
 	cairo_surface_t *test_surface;
 	cairo_t *test_cairo; // used to estimate font/text sizes
 	enum auth_state auth_state;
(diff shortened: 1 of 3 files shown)


kennylevinsen (approved):
LGTM, thanks!

PR #294, opened 2023-05-03
Separate password validation and input state machines

mstoeckl:
This commit mostly decouples the parts of the indicator state related to authentication messages from those related to keyboard input. This makes it possible to type input while a previous password is verifying, without hiding the `verifying/wrong` message as soon as a key is pressed.

Note: i3lock currently has the ring highlight indicators vanish about 0.25 seconds after a key press; since Swaylock currently keeps them around until transitioning to the idle state, I have not tried to match the i3lock behavior more closely here.

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -8,14 +8,20 @@
 #include "pool-buffer.h"
 #include "seat.h"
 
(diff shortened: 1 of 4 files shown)


emersion (approved):
Thanks!

PR #301, opened 2023-06-29
Show cleared state when backspacing last character

iyzana:
Typing `A<backspace>` did not result in the cleared state even though the buffer was empty. One had to press backspace one more time to reach it.

I also tried always returning `pw->len != 0` from `backspace()` after the buffer has been modified, which fixes this as well.
But I decided against that, because then the meaning of the return value is unintuitive.


emersion (approved):
Thanks!

Issue #312, opened 2023-09-04
Fails to build with Clang due to -Werror=sign-compare


jbeich:
Regresed by db9ee6d1277b as CI doesn't cover Clang or FreeBSD. From [error log](https://github.com/swaywm/swaylock/files/12516128/swaylock-1.7.2.8.log):

```c
$ export CC=clang CXX=clang++
$ meson setup _build
$ meson compile -C _build
[...]
../main.c:1257:64: error: comparison of integers of different signs: 'ssize_t' (aka 'long') and 'unsigned long' [-Werror,-Wsign-compare]
                if (write(state.args.ready_fd, ready_str, strlen(ready_str)) != strlen(ready_str)) {
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~
```

kennylevinsen:
In discussions related to seatd, we also suspect that this mechanism - trying to satisfy systemd's notify protocol on a passed fd - might unfortunately be flawed.

So we might want to just revert to purely s6 readiness (writing a single '\n') while we're at it.

PR #317, opened 2023-10-05
Don't send READY=1 for readiness notifications

emersion:
Just send a singular newline like s6 expects.

systemd doesn't support spawning a process with an FD to send readiness notifications to, instead it provides a socket name. IOW, this cannot be used directly with systemd after all.

Closes: https://github.com/swaywm/swaylock/issues/312


kennylevinsen (approved):
LGTM

PR #283, opened 2023-01-29
Fingerprint support

SL-RU:
I've added fingerprint support via fprintd daemon which has DBus communication and because of that build dependencies was added(`dbus`, `gio`, `fprintd` and etc), but they are not required for execution. To run swaylock with fprintd you need to start fprintd and then swaylock with argument `-p` or `--fingerprint`

Can you make a review and say is merge possible?

This relates to #61

PR #324, opened 2023-11-08
Implement handling SIGUSR2 to record external failed authentication attempts

lilyinstarlight:
This adds support for indicating external authentication failures using `SIGUSR2`, in addition to the existing support for external authentication successes using `SIGUSR1`

Since `SIGUSR*` defaults to termination, we ignore them in the forked pw backend children (which worked already because `SIGUSR1` ends up in termination anyway, but we do not want that to happen for `SIGUSR2`)

This is intended to be usable with an external program that can handle fprintd communication as an alternative to #283

Supersedes and closes #311

### DIFF
--- a/comm.c
+++ b/comm.c
@@ -1,3 +1,5 @@
+#define _POSIX_C_SOURCE 200809L
+#include <signal.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <sys/types.h>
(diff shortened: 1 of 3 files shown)


emersion:
SIGUSR1 is not intended for external processes to indicate success. It's a feature to unlock the session when it's stuck.

lilyinstarlight:
> SIGUSR1 is not intended for external processes to indicate success. It's a feature to unlock the session when it's stuck.

Adding a SIGUSR2 was suggested in https://github.com/swaywm/swaylock/pull/283#issuecomment-1678615442. If this is not the correct approach, is there something else that can be done to support this use-case?

Thanks!

emersion:
swaylock is not meant to be extensible through plugins. I don't think we want to support extensibility because of the complexity. Regardless, signals in particular are not a great API.

I would suggest forking swaylock or creating your own screen locker with fprintd support.

lilyinstarlight:
Okay, feel free to close this then. Thanks!

PR #327, opened 2023-11-26
Link to manpage

lhanson:
The README for `swayidle` has a convenient link to the manpage, following suit here.


emersion (approved):
Thanks!

PR #330, opened 2023-12-06
Updated: Fix output-specific images when output reappears

alexbakker:
Since #155 hasn't been touched in quite a while, I figured I'd take a shot at dragging it across the finish line. The patch has been rebased and modified to use wl_output instead of xdg_output. The remaining feedback from the previous pull request has also been addressed.

### DIFF
--- a/main.c
+++ b/main.c
@@ -152,6 +152,8 @@ static void create_surface(struct swaylock_surface *surface) {
 		wl_surface_set_opaque_region(surface->surface, region);
 		wl_region_destroy(region);
 	}
+
+	surface->created = true;
 }
 
 static void ext_session_lock_surface_v1_handle_configure(void *data,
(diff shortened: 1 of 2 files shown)


emersion:
Hm, this looks a bit fragile. I would suggest moving this

https://github.com/swaywm/swaylock/blob/f2298bdbf788678a9e08fd17a3e37600767f5a9b/main.c#L314

to `handle_wl_output_done` instead.

emersion:
Would need to add a check to make sure `create_surface()` only runs once per surface.

alexbakker:
Alright. How's this?

main.c
```diff
@@ -234,7 +236,11 @@ static void handle_wl_output_mode(void *data, struct wl_output *output,
 }
 
 static void handle_wl_output_done(void *data, struct wl_output *output) {
-	// Who cares
+	struct swaylock_surface *surface = data;
+	if (!surface->created && surface->state->run_display) {
+		create_surface(surface);
+		wl_display_roundtrip(surface->state->display);
```
emersion, line 242: I don't think this is necessary (in fact, I think it's a potential source of bugs), but it was there before so orthogonal to this PR.

emersion (approved):
LGTM, thanks!

main.c
emersion: https://github.com/swaywm/swaylock/pull/336

PR #331, opened 2023-12-09
Clear password on ctrl+backpace and ctrl+delete

### LINKED ISSUE #157: Clear password input on Ctrl+Backspace
See the title. Some users would also expect input to clear itself after holding backspace for some time (e.g. 600ms).

prezmop:
See title. This is a common convention in many programs that have text entry fields.

Resolves #157.


emersion (approved):
Thanks!

PR #339, opened 2024-01-16
Install pam config only if pam is enabled

Sertonix:


Sertonix:
Just saw that #325 includes the same change. So if that is merged this isn't needed anymore.

Issue #175, opened 2021-04-27
setuid wrongly required with TCB shadow


lanodan:
```
0 $ swaylock
2021-04-27 13:10:07 - [swaylock-1.5/shadow.c:19] swaylock needs to be setuid to read /etc/shadow
1 $ ls -lah /etc/tcb/$(id -un)/shadow
-rw------- 1 haelwenn shadow 135 Apr 13 11:50 /etc/tcb/haelwenn/shadow
0 $ doas stat /etc/shadow
stat: cannot statx '/etc/shadow': No such file or directory
0 $ grep '^shadow:' /etc/nsswitch.conf 
shadow:     tcb compat
0 $
```

This is described at <https://www.openwall.com/tcb/>, it is supported natively in musl libc and otherwise via `libnss_tcb.so` as configured for me in `/etc/nsswitch.conf`.
It's probably also not required with other alternative authentication methods and looking into `shadow.c` there is already code to handle fails in `getpwuid` & `getspnam`.

emersion:
Patches welcome.

sertonix:
Does it work if you remove this safety check? https://github.com/swaywm/swaylock/blob/91bb96811054e66022b109e269a4fabaa2221e2b/shadow.c#L18-L22

PR #340, opened 2024-01-16
Read password hash before fork

sertonix:
This ensures that the parent properly errors only if the password cannot be read.

### DIFF
--- a/shadow.c
+++ b/shadow.c
@@ -1,4 +1,5 @@
 #define _XOPEN_SOURCE // for crypt
+#include <assert.h>
 #include <pwd.h>
 #include <shadow.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


sertonix:
This might introduce more issues similar to !175. I think a lot more issues would be fixed by removing the safety check.

shadow.c
```diff
@@ -15,11 +16,15 @@
 #include "swaylock.h"
 
 void initialize_pw_backend(int argc, char **argv) {
-	if (geteuid() != 0) {
+	int fd_shadow = open("/etc/shadow", O_CLOEXEC | O_RDONLY);
```
kennylevinsen, line 19: If you do not want a file descriptor, consider using `access(2)` instead to just check the file permissions.

kennylevinsen:
This will not work for #175, where the shadow file is located elsewhere. For that, the checks would have to be replaced with a password retrieval attempt (the start of `run_pw_backend_child`).

sertonix:
> This will not work for #175, where the shadow file is located elsewhere. For that, the checks would have to be replaced with a password retrieval attempt (the start of `run_pw_backend_child`).

I looked at the code and found a better solution. Will push as soon as it is done.

sertonix:
Is it safe to pass a pointer to a fork? This code does so.

emersion:
A downside of this approach is that changing the user's password while swaylock is running won't work. In other words, if someone forgets their password and tries to change it via SSH, swaylock still won't unlock. Maybe that's not such a big deal since it's possible to send a signal to swaylock to unlock itself via SSH too.

sertonix:
> A downside of this approach is that changing the user's password while swaylock is running won't work.

I am pretty sure that is already how it works even without my change.

emersion:
Hm, indeed!

sertonix:
The privilege drop would also not work with reading the password later.

sertonix:
@emersion What is the blocker for this PR?

emersion (approved):
Thanks!

PR #341, opened 2024-01-16
Check setgid too after dropping root

sertonix:


shadow.c
```diff
@@ -33,9 +33,9 @@ void initialize_pw_backend(int argc, char **argv) {
 		swaylock_log_errno(LOG_ERROR, "Unable to drop root");
 		exit(EXIT_FAILURE);
 	}
-	if (setuid(0) != -1) {
+	if (setuid(0) != -1 || setgid(0) != -1) {
```
kennylevinsen, line 36: nit: The we handle gid and uid separately above. No opinion as to whether it should be one or two if's, but consistency is nice. :)

sertonix: So you mean creating a second if statement for setgid?

emersion: I think it's fine. Thanks!

Issue #346, opened 2024-02-22
All `finished` events are resulting in an `exit(2)`, when the protocol states otherwise


mattkae:
I noticed that whenever a `finished` event is sent to `ext_session_lock_v1`, swaylock exits with `2`. According to the spec:

>  If the locked event is sent on creation of this object the finished event may still be sent at some later time in this object's lifetime. This is compositor policy.
>  Upon receiving this event, the client should make either the destroy request or the unlock_and_destroy request, depending on whether or not the locked event was received on this object.

In this case, I would expect `swaylock` to send an `unlock_and_destroy` request and exit gracefully.

**Use case**: This could happen because the compositor's policy accepts unlocking via logind (or some other 3rd party mechanism) as a valid unlock. In this case, we want the compositor to inform the client that they can exit without any issues.

Let me know if I'm misunderstanding something here. Cheers!

emersion:
I'm not sure this is a good thing to do. I'm worried about unlocking in other situations that when the user has correctly typed their password. If the compositor decides to unlock, they shouldn't care that the client sends the `unlock_and_destroy` request.

kennylevinsen:
I suppose this issue is primarily about the exit code. To sort that part out, we could `exit(0)` if we have previously received confirmation of lock, and `exit(2)` otherwise, so scripts that called swaylock can see that it was not a failure.

(Sending `destroy` before existing is redundant, the compositor will clean up.)

mattkae:
An `exit(0)` would solve my particular problem. However, I see why an `exit(2)` was being used, because this `finished` event is sent both when the compositor is done with the lock _and_ when the compositor denies the lock altogether. I would expect an `exit(0)` in the former and an `exit(2)` in the latter, as the latter means the lockscreen didn't work at all.

emersion:
I'm a little uneasy exiting with a zero code when the user hasn't explicitly unlocked the screen by typing the correct password.

mattkae:
Yeah I understand that. The fact that the `finished` event is coupled with a `denied` event isn't helping matters either. The spec says "This might also be sent because the compositor implements some alternative, secure way to authenticate and unlock the session" but how could we know from the `finished` event whether or not that was sent in response to a failure.

If that's the case, I'd be okay leaving it for the time being. We might eventually require an updated spec if we're serious about the idea that compositor may implement another avenue for authentication (e.g. logind) :wink:

PR #347, opened 2024-03-04
Exiting with status 2 only if we actually failed to lock

mattkae:
This is a suggested fix for: #346 

Because we can distinguish between a failure to lock the screen and a simple request from the compositor to "go away now", we can `exit(2)` in the former case and `exit(0)` in the latter.

### DIFF
--- a/main.c
+++ b/main.c
@@ -277,9 +277,14 @@ static void ext_session_lock_v1_handle_locked(void *data, struct ext_session_loc
 }
 
 static void ext_session_lock_v1_handle_finished(void *data, struct ext_session_lock_v1 *lock) {
-	swaylock_log(LOG_ERROR, "Failed to lock session -- "
-			"is another lockscreen running?");
-	exit(2);
+	struct swaylock_state *state = data;
+	if (state->locked) {
+		exit(0);
+	} else {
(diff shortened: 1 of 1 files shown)


emersion:
I would prefer to keep 0 for "user unlocked" and pick something else for "compositor unlocked".

mattkae:
> I would prefer to keep 0 for "user unlocked" and pick something else for "compositor unlocked".

This depends on whether or not the compositor telling the lockscreen to go away is an error case. In my opinion, it isn't because the compositor may internally decide that it's time to unlock, which means that it wants to relate that info to the client.

emersion:
No, it's a question of the semantics or the exit code. To me, it's a good design to specify that 0 is returned if and only if credentials were successfully verified. No credentials are verified when the compositor decides to close the lock screen.

AlanGriffiths:
The question is what are the most useful semantics for the exit code. Usually, zero indicates success and non-zero indicates a failure. Being dismissed by the compositor is not an error.

Note that the case that is motivating @mattkae is when the credentials are validated elsewhere, e.g. by the greeter, which notifies the compositor and the compositor notifies the lockscreen (by sending "finished"). In this scenario swaylock  indicates an error when no error has occurred.

PR #353, opened 2024-03-30
Use explicit_bzero(3) to clear passwords, if it is available

nmeum:
This is a revived version of #148 which is less invasive by only modifying the implementation of `clear_buffer()`.

Relying on the volatile keyword to clear memory is, in many cases, insufficient and may result in the clearing operation being optimized out by the compiler. Recognizing the difficulty of zero'ing memory, many libcs provide an explicit_bzero(3) function (which the libc guarantees not to be optimized out). This is presently implemented by OpenBSD, musl libc and glibc. This patch makes swaylock use this function and, if it is not available, falls back to the existing code and prints a warning during build.

For more details on the complexity of zero'ing memory, refer to the following 35c3 talk: https://media.ccc.de/v/35c3-9788-memsad

### DIFF
--- a/password.c
+++ b/password.c
@@ -1,3 +1,4 @@
+#define _BSD_SOURCE // for explicit_bzero
 #include <assert.h>
 #include <errno.h>
 #include <pwd.h>
@@ -13,12 +14,16 @@
 #include "unicode.h"
 
 void clear_buffer(char *buf, size_t size) {
+#ifdef HAVE_EXPLICIT_BZERO
+	explicit_bzero(buf, size);
+#else
 	// Use volatile keyword so so compiler can't optimize this out.
 	volatile char *buffer = buf;
 	volatile char zero = '\0';
 	for (size_t i = 0; i < size; ++i) {
 		buffer[i] = zero;
 	}
+#endif
 }
 
 void clear_password_buffer(struct swaylock_password *pw) {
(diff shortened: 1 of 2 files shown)


password.c
```diff
@@ -1,3 +1,4 @@
+#define _BSD_SOURCE // for explicit_bzero
```
nmeum, line 1: Defining feature tests macro in the source file is somewhat unideal, only doing it here because that's what is also done elsewhere in the code. Refer to [this post](https://ariadne.space/2021/12/21/stop-defining-feature-test-macros-in-your-code/) for more information.

michaelortmann:
you could also opt to use a more generic sodution to clear_buffer().
i can recommend to just copy the code from here:
https://github.com/jedisct1/libsodium/blob/master/src/libsodium/sodium/utils.c#L125-L157

nmeum:
The libsodium compat code is also recommended in the referenced talk. However, given the prior discussion in #148 and the comments regarding non-posix functions, I assumed that something like this is even more less likely to get merged. My goal here is to propose a more minimal integration, in the hopes that this will be considered for inclusion. Also recall that `explicit_bzero` should nowadays be supported by all major platforms targeted by swaylock, hence I am not sure if additional fallbacks are even needed these days.

emersion:
Note, the standard equivalent of this is C23's `memset_explicit`. However, it doesn't seem implemented in libcs yet.

nmeum:
Note that `memset_explicit` has only [recently been standardized with C23](https://www.open-std.org/JTC1/SC22/WG14/www/docs/n2897.htm), hence it will so it will be a while before it is widely available.

Issue #357, opened 2024-04-29
Image upside down (fixed on swaybg)


enoryw:
There's an update to swaybg that [fixed some images not orientated correctly](https://github.com/swaywm/swaybg/pull/68). I assume swaylock can also pull this fix? A release would also be much appreciated. Previously I manipulated some images for workaround.

Apologies if it's not related.

emersion:
Would you be willing to send a similar patch?

PR #361, opened 2024-06-14
Correct for image orientation when loading image (fix from swaybg)

drnlm:
JPEG and other image formats may include an EXIF orientation tag which indicates in what orientation (rotation and mirroring) the image should be displayed. libgdk-pixbuf does not correct for this when loading an image, but provides a function to apply the transform after the fact, which this commit uses.

Fix pulled from swaybg - see https://github.com/swaywm/swaybg/pull/68 for the original patch

Closes: #357

### DIFF
--- a/background-image.c
+++ b/background-image.c
@@ -31,8 +31,12 @@ cairo_surface_t *load_background_image(const char *path) {
 				err->message);
 		return NULL;
 	}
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

rieje:
Nice! Time for a release?

PR #368, opened 2024-08-01
Support output IDs

wille:
I need to support targeting the output with the output identifier (`Dell Inc. DELL U3223QE GB6Y9P3`) rather than just the output name (eg. `DP-6`) because the output name might change every time I replug my monitors and I want my internal OLED display to have no image while my external monitors should have a nice image.

Example:
`swaylock -f -c 000000 --image "Dell Inc. DELL U3223QE GB6Y9P3":~/Pictures/1.jpg -d`

The wl_output_description returns the full model, make and output name so I am trimming the output name from the output description to match kanshi and sway output id matching.

### DIFF
--- a/include/swaylock.h
+++ b/include/swaylock.h
@@ -118,6 +118,7 @@ struct swaylock_surface {
 	int32_t scale;
 	enum wl_output_subpixel subpixel;
 	char *output_name;
+	char *output_description;
 	struct wl_list link;
 	// Dimensions of last wl_buffer committed to background surface
 	int last_buffer_width, last_buffer_height;
@@ -127,6 +128,7 @@ struct swaylock_surface {
 struct swaylock_image {
 	char *path;
 	char *output_name;
+	char *output_id;
 	cairo_surface_t *cairo_surface;
 	struct wl_list link;
 };
--- a/main.c
+++ b/main.c
@@ -259,7 +259,19 @@ static void handle_wl_output_name(void *data, struct wl_output *output,
 
 static void handle_wl_output_description(void *data, struct wl_output *output,
 		const char *description) {
-	// Who cares
+	swaylock_log(LOG_DEBUG, "handle_wl_output_description %s ", description);
+
+	struct swaylock_surface *surface = data;
+
+	// The output description will be in the format "<make> <model> <serial> (<output>)" like Dell Inc. DELL U3223QE G76Y9P3 (DP-5)
+	// so cutting it at the last space
+	char *separator = strrchr(description, ' ');
+	if (separator) {
+		*separator = '\0';
+		surface->output_description  = strdup(description);
+	}
+
+	swaylock_log(LOG_DEBUG, "set _%s_", surface->output_description );
 }
 
 struct wl_output_listener _wl_output_listener = {
@@ -349,7 +361,7 @@ static cairo_surface_t *select_image(struct swaylock_state *state,
 	struct swaylock_image *image;
 	cairo_surface_t *default_image = NULL;
 	wl_list_for_each(image, &state->images, link) {
-		if (lenient_strcmp(image->output_name, surface->output_name) == 0) {
+		if (lenient_strcmp(image->output_name, surface->output_name) == 0 || lenient_strcmp(image->output_name, surface->output_description) == 0) {
 			return image->cairo_surface;
 		} else if (!image->output_name) {
 			default_image = image->cairo_surface;


emersion:
Unfortunately, this adds assumptions about the format of the output description which are unreliable and non-portable across compositors. I'd rather not do that.

wille:
You're definitively right, I was wondering if you can point me in the right direction to reliably get this ID just as in Kanshi or sway `output`

emersion:
Sadly there is no way to get this information. I'm not sure just adding an event to pass it to clients is a good way forward, an opaque tag might be better, see the discussions in https://gitlab.freedesktop.org/wayland/wayland/-/merge_requests/109#note_970789

wille:
Sorry I don't follow, both sway and kanshi supports targeting the output by the identifier which I want to do here as well.

I just pushed a new commit to simply check if the output description contains the output name provided to --image, so `--image "Dell Inc. DELL U3223QE GB6Y9P3":~/Pictures/1.jpg` would match the output_descripton `Dell Inc. DELL U3223QE GB6Y9P3 (DP-X)`

trinitronx:
From the perspective of user-friendly per-display configuration, the use-case behind this PR is a great idea.  It would unify the configuration matching behavior of output specifiers for `swaylock` with `sway`, `swaybar`, `waybar`, `kanshi` and similar `wlroots`-based projects.  Additionally, `wlroots`-based compositors including `sway`, and components like `swaybar` already support using the EDID-based identifier (e.g. `make` + '` `' + `model` + '` `' `serial`) as in swaywm/sway@50d36ef23, [`kanshi`][0], [wlroots DRM backend][1], [wlroots `wlr_output` struct][2], and of course the [`wlr-output-management-unstable` protocol][3] [since version 2 supports `make`, `model` and `serial`][4].

The nice thing these days is that EDID is well supported by the vast majority of output displays, thanks to the standardization efforts of [VESA][5], and [CTA (formerly CEA)][6] we have [CTA-861][6.1].  For the edge cases where devices deviate from the standards or misbehave, the EDID data [can be overridden by placing files in `/lib/firmware` with the Linux Kernel's `CONFIG_DRM_LOAD_EDID_FIRMWARE` option][7], or in the case of supported `v4l2` devices, the [`v4l2-ctl` utility can be used to set it via `VIDIOC_S_EDID`][8].  One notable example of the latter option `v4l2-ctl` is [used by PiKVM][9] to set the EDID from a file managed by the [`kvmd-edidconf` utility][10].

Another nice thing is that [EDID is also supported by the vast majority of DRM drivers][11].  Might be worth considering using at least the `make`, `model`, and `serial` for the user experience benefits of identifying a display in configuration files.  It's generally unique _enough_ for most cases to identify a display output in a port-agnostic way.  In those cases where it doesn't, the user can override the EDID data using aforementioned methods, or simply fallback to the old name matching and use the port identifiers (e.g. `HDMI-A-N`, `DP-N`) and plug in the displays accordingly.  However, the benefits of EDID-based data are great for identifying a particular display especially for devices with hot-pluggable displays (e.g. laptops) and users who often plug into the same displays frequently.  Examples might be plugging into a particular make & model of projector in meeting rooms, plugging in an HDMI output for temporary extended displays, etc...  Being able to have a "public-facing" `swaylock` background image for the external displays would be a great benefit.

[0]: https://git.sr.ht/~emersion/kanshi/tree/master/item/main.c#L26-31
[1]: https://github.com/swaywm/wlroots/blob/eeb7a8113869790e5f6fc643bafa577b08b5f4c3/backend/drm/drm.c#L1213-L1222
[2]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_output.h#L160-161
[3]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/protocol/wlr-output-management-unstable-v1.xml#L263-331
[4]: https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/types/wlr_output_management_v1.c#L854-862
[5]: https://vesa.org/
[6]: https://web.archive.org/web/20221011163310/https://www.cta.tech/Resources/i3-Magazine/i3-Issues/2019/November-December/cta-861-ctas-most-popular-standard
[6.1]: https://en.wikipedia.org/wiki/Extended_Display_Identification_Data#CEA-861-E
[7]: https://www.kernel.org/doc/html/v5.8/admin-guide/edid.html
[8]: https://github.com/cz172638/v4l-utils/blob/7b2d48ff594dcc2c9b395463441e5abb4f5e9439/utils/v4l2-ctl/v4l2-ctl-edid.cpp#L1303
[9]: https://github.com/pikvm/kvmd/blob/85a2f2367d3c7e29f6e1857e110d47dfd5297814/configs/os/services/kvmd-tc358743.service#L9
[10]: https://docs.pikvm.org/edid/
[11]: https://drmdb.emersion.fr/properties/3233857728/EDID

Issue #370, opened 2024-08-04
swaylock after swayidle does `dpms off` and starts swaylock causes system to freeze when swayidle invokes `dpms on` for resume


Stephen-Seo:
I'm not sure where the right place for this issue is, but here's the details.

On ArchLinux with `sway`, `wlroots`, `swayidle`, and `swaylock` packages in the official repositories, there is no issue.

But if `sway-git` and `wlroots-git` is installed from the AUR, the swaylock instance started by swayidle causes a system freeze on resume when swayidle does `dpms on` on resume. The system becomes unresponsive, one cannot switch to another tty, and the power button is ignored when by default it should shutdown the system.

Note I also have `libdisplay-info-git` and `xorg-xserver-git` installed from AUR as well, which is built before I build `wlroots-git`.

I have tried `swaylock-git` and `swayidle-git` and the issue still occurs with `sway-git`.

Nothing is logged to systemd-journald, and the output of sway also does not have any logs when the freeze occurs.

My swayidle command is this:

    exec swayidle -w \
             timeout 780 'swaylock -f -c 007050 -i /usr/share/backgrounds/archlinux/geolanes.png' \
             timeout 600 'swaymsg "output * dpms off"' resume 'swaymsg "output * dpms on"' \
             before-sleep 'swaylock -f -c 007050 -i /usr/share/backgrounds/archlinux/geolanes.png'

Stephen-Seo:
I've been doing some testing with this issue, and tried running `sway -V` and `swayidle -d`. What shows up when sway's stdout and stderr is piped into a file after a freeze is this:

    Aug 04 15:05:47 2024-08-04 15:05:47 - [Line 650] idle state
    Aug 04 15:05:47 2024-08-04 15:05:47 - [Line 146] Cmd exec swaylock -f -c 007050 -i /usr/share/backgrounds/archlinux/geolanes.png
    Aug 04 15:05:47 2024-08-04 15:05:47 - [Line 172] Spawned process swaylock -f -c 007050 -i /usr/share/backgrounds/archlinux/geolanes.png
    Aug 04 15:05:47 2024-08-04 15:05:47 - [Line 174] Blocking until process exits
    Aug 04 15:05:47 2024-08-04 15:05:47 - [Line 179] Process exit status: 0
    Aug 04 15:05:49 00:02:47.218 [INFO] [wlr] [backend/drm/drm.c:1672] Scanning DRM connector 128 on /dev/dri/card1
[...]

PR #371, opened 2024-09-06
Avoid getting render-loop stuck

kennylevinsen:
One could quite consistently get the render loop stuck by waking the system with a keypress, such that swaylock would receive a configure event shortly before rendering a new frame.

Reproduction script:
```sh
swayidle \
        timeout 3 'swaymsg "output * power off"' resume 'swaymsg "output * power on"' \
        &

swaylock -e -F
kill %swayidle
swaymsg "output * power on"
```

Run that, then wake the machine with backspace. swaylock renders the "Cleared" text, and nothing more. It can be a bit timing sensitive.

Fixes: https://github.com/swaywm/swaylock/issues/370

### DIFF
--- a/render.c
+++ b/render.c
@@ -33,7 +33,23 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
 	}
 }
 
-void render_frame_background(struct swaylock_surface *surface) {
+static void surface_frame_handle_done(void *data, struct wl_callback *callback,
+		uint32_t time) {
+	struct swaylock_surface *surface = data;
+
+	wl_callback_destroy(callback);
+	surface->frame = NULL;
+
+	render(surface);
+}
+
+static const struct wl_callback_listener surface_frame_listener = {
+	.done = surface_frame_handle_done,
+};
+
+static bool render_frame(struct swaylock_surface *surface);
+
+void render(struct swaylock_surface *surface) {
 	struct swaylock_state *state = surface->state;
 
 	int buffer_width = surface->width * surface->scale;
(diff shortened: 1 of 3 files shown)


render.c
```diff
@@ -33,7 +33,23 @@ static void set_color_for_state(cairo_t *cairo, struct swaylock_state *state,
...
 
-void render_frame_background(struct swaylock_surface *surface) {
+static void surface_frame_handle_done(void *data, struct wl_callback *callback,
+		uint32_t time) {
+	struct swaylock_surface *surface = data;
+
+	wl_callback_destroy(callback);
+	surface->frame = NULL;
+
+	render(surface);
```
Nefsen402, line 43: Why are we rendering straight away after we handle the frame done? Wouldn't we render a frame, commit it, wait for the done event, then just immediately render again?

Nefsen402 (commented):
Not sure how this is supposed to work, it appears that all the old damage code was replaced with immediately rendering the frame.

render.c
kennylevinsen: `render` checks at the beginning if the surface is dirty and bails if that is not the case, stopping the loop.

My original intent was to move all the commits *out* of render, but render (before: render_frame_background) makes a temporary buffer that it destroys, so commit has to happen within that function. Going the other way and having it absorb the callback was a quicker solution.

Nefsen402: I see. The names are a bit confusing being `render` and `render_frame` but that can be cleaned up later.

Nefsen402 (approved):
Thanks!

PR #375, opened 2024-09-19
Only submit password if previous validation completed

### LINKED ISSUE #372: Holding Enter key (for example while handling keyboard) queues large number of unlock attempts
Each unlock attempt takes quite a while, and with many queued the screen effectively cannot be unlocked for a long time.

There is an option to ignore empty passwords which would likely be a workaround if it is known that the password cannot be empty.

However, there are situations in which it is desirable to support both empty password and locking. For example, when building a live medium it is desirable to build it with an empty password so that users do not have to always ask around how to log in, and at the same time support securely locking the system if the user sets a password after the system is booted.

A solution would be to ignore input typed until the verification attempt finishes, or to special-case empty password, and ignore it once it fails to unlock the system. The latter can be prone to problems if the password is only one factor, and other factor is configured in PAM.

mstoeckl:
This should resolve #372; while I'm not entirely satisfied with this solution, I think it is a net improvement.

Commit message:
> This averts the risk that one can queue up many password submissions by pressing enter repeatedly (or with key repeat, holding enter), much more quickly than PAM will process them. It may also improve usability by reducing unrevealed system state: now that at most one password can be queued, the "Verifying" message is always shown when a password is being validated.

> However, this commit does introduce a minor UI issue: if one types an invalid second password just slightly faster than PAM validation completes, it may fail to submit, and the signs of this may be hard to notice (state transition timing, presence of keypress indicator, and (if enabled) attempt count).

### DIFF
--- a/password.c
+++ b/password.c
@@ -111,6 +111,9 @@ static void submit_password(struct swaylock_state *state) {
 	if (state->args.ignore_empty && state->password.len == 0) {
 		return;
 	}
+	if (state->auth_state == AUTH_STATE_VALIDATING) {
+		return;
+	}
 
 	state->input_state = INPUT_STATE_IDLE;
 	state->auth_state = AUTH_STATE_VALIDATING;


emersion (approved):
Thanks!

Issue #377, opened 2024-10-02
Unlocking via Yubikey regression since updating to 1.8.0


b0ch3nski:
Hi,

I've noticed little regression after updating to [1.8.0](https://github.com/swaywm/swaylock/releases/tag/v1.8.0).

Behavior on 1.7.2:
1. Screen is locked, I'm pressing "Enter"
2. Yubikey blinks, after touching Yubikey screen unlocks, Yubikey stops blinking

Behavior on 1.8.0:
1. Screen is locked, I'm pressing "Enter"
2. Yubikey blinks, after touching Yubikey screen unlocks, but Yubikey still blinks (or starts blinking again? hard to notice) like it's waiting for another touch input

Here is a relevant `/etc/pam.d/system-auth` section:
```
#%PAM-1.0
auth       sufficient                  pam_u2f.so           authfile=/etc/u2f_mappings cue
auth       required                    pam_faillock.so      preauth
...
```

My guess is that for some reason, swaylock tries to go through PAM twice ... ?

mstoeckl:
I don't have the setup to reproduce this issue, but suspect it's caused by swaylock queuing multiple unlock requests to its PAM handling subprocess, which tries to process all of them instead of stopping at the first success. (See [`run_pw_backend_child()`](https://github.com/swaywm/swaylock/blob/cca2436ba535d17c49b7d7ba75ebfa6a18ec9209/pam.c#L103).) If my theory is correct, then building swaylock using either of #379 and #375 should prevent the described behavior.

b0ch3nski:
@mstoeckl You must have been right - issue is not present in 1.8.2 🎉

PR #379, opened 2024-10-12
Stop processing auth requests after success

mstoeckl:
The PAM handling subprocess can receive unsuccessful requests after a successful one. (For example: press enter a few times to build up a queue of empty password attempts, enter a correct password, and then press enter a few times to queue a few more empty password attempts.) Processing the trailing requests is unnecessary, and might cause side effects like those observed in  https://github.com/swaywm/swaylock/issues/377.


emersion (approved):
Thanks!

PR #381, opened 2024-11-01
Exit when password handling subprocess crashes

mstoeckl:
This change makes swaylock exit immediately when the pam child process crashes or exits early. (Once the child process has crashed, it is no longer possible to unlock the screen; having swaylock exit is now safe because ext-session-lock-v1 is required, and is better than the alternatives of operating in a state where future unlock attempts all fail, or crashing.)

To test this, run swaylock, identify its child process, and `kill` it.


emersion (approved):
LGTM, thanks!

PR #387, opened 2024-11-17
Fix potential NULL deference after getpwuid

OkamiW:
This can happen if swaylock is mistakenly run inside a sandbox without /etc/passwd


emersion (approved):
Thanks!

Issue #270, opened 2022-12-14
Support 2FA pin code and password with security keys


NgoHuy:
Hi teams,

Thank for great application, I have use yubikey and setup login requires pin and password, but after I enter the pin, it sent to pam quickly, not required password.

The pam's setup is here:
```
$ cat /etc/pam.d/system-auth
#%PAM-1.0

auth      required pam_u2f.so pinverification=1 cue
#auth      sufficient pam_fprintd.so
auth      required  pam_unix.so     try_first_pass nullok
```

I think it should handle seperate pam module instead send pam quickly as enter first pass.

notmentaloutlaw:
You can already do this just modify /etc/pam.d/swaylock just like you did with system-auth

NgoHuy:
> You can already do this just modify /etc/pam.d/swaylock just like you did with system-auth

I did it before submitting the issue, it used first module then pass to pam handle

Issue #244, opened 2022-06-08
Design: Handle complex PAM authentifications


VannTen:
Hello,

I started to play with my Yubikeys recently and would like to work on swaylock
so it can handle the UI part a bit better.

(The objective would be to fix #61 and #213)

I'm not super familiar with the codebase so I wanted to check with you before
starting to code.

As I understand, currently, swaylock handle the auth process in a separate
process, which communicates with the main one (doing rendering, event looping
etc) using pipes. (in pam.c and comm.c). The main process pass the password
string and the "auth" process respond with ok/ko (it's a simplification)

In order to support "richer" conversation with PAM, we need:

1. A way to send back a string from the "auth" process to the main one, so it
   can displays prompts from PAM.
2. Displaying that string, so probably a new surface in the rendering part.
3. Some modification the control flow in the "auth" process.

My first question, should I re-use the same pipe used to communicate the
password string to the child process to send back prompt string, or should I use
a dedicated one, something else altogether ? (Is there a particular policy on
how the password is handled, or something like that ?)

Secondly, I think the control flow should be driven by the code in pam.c,
particularly in the `handle_converstation` fonction. The initial state of
swaylock (requiring a password, prompting user to insert a token, etc) will
depend on the particular pam configuration and what msg_style pam has sent us.
That seems like a big change so would it be ok ?

I think we need two new states (for lack of a better word):
1. prompting the user for action, then confirm (ex: insert u2f keys, press
   ENTER)
2. prompting the user for action, pass control to PAM (ex: touch your u2f key).

Those two could be assimilated, to, respectively, the AUTH_SPACE_INPUT (writing
the password) and AUTH_STATE_VALIDATING ("verifying") (which is mostly what's
going on currently I think) but it's not a very clear UX.

I'm only testing with a Yubikey (using pam_u2f) so I'm only considering those
states/action right now but if you see other cases don't hesitate to mention it.

----

I would very much appreciate feedback on this very rough sketch, or things you
think I should think about, or questions if I'm not clear.

NgoHuy:
I have another case, if I used `auth required` bose yubikey and password, swaylock never asks for password, only once for yubikey pin. I tested on KDE, it's same as swaylock.

hramrach:
So basically you have multiple passwords. That generally falls into the 'control flow should be driven by pam' as in if and when a password is requested depends on pam configuration, and the hardcoded flow to send one password and wait for something to happen is not suited for all cases.

[...]

Issue #213, opened 2021-10-14
[enhancement] Forward text printed by PAM


chiraag-nataraj:
As part of my general security-conscious setup, I have enabled `libpam-u2f` so that I get prompted to touch my Yubikey after I enter my password (as a second factor). I have enabled the `cue` option just for additional feedback, which works well on the terminal (it prints "Please touch the device." by default, but that's configurable using the `cue_prompt` option). However, when I lock the screen with `swaylock`, I simply get a `Verifying` text and the `cue` prompt never shows up.

It would be nice if this (and any other text printed out by PAM during authentication) were directly forwarded to the lock screen (potentially as a configurable option for those who don't want it). Ideally, any additional input (à la `libpam-otpw`) would also be captured and forwarded like the original password.

hendursaga:
I could test any patches with my Solo key (v2) once it arrives!

Issue #61, opened 2019-03-19
Have to scan fingerprint after entering password


JustMeErazem:
Swaylock version: 1.3
OS: Arch Linux
Sway version: 1.0
Laptop: Thinkpad T540p

If I lock the screen with swaylock I have to enter the password (obviously), but after entering the correct password, it will not let me in. This is when my fingerprint reader starts showing its "scanning" light. It only lets me in when I scan my fingerprint.

The scanner doesn't start until I enter something (Not necessarily my actual password - can even press Enter) and doesn't work even if entering right password (until I scan my fingerprint).

rsansores:
I'm not sure if this is a bug. What is happening is that PAM is launched when you press "Return" key. Your PAM must be configurated to first request fingerprint so even if you write your password PAM will ignore the input and wait for the fingerprint.

In a GDM this is not an issue because there are two steps.

Step 1: request the user to login
Step 2: request password

After step 1 PAM is called. But in the end, sway-lock is not a GDM, is just a tool to lock a wayland session.

As workaround you could:

- Configure PAM to not request fingerprint for swaylock.

or

- Unlock without password. Just press enter to unlock and put your finger in the fingerprint reader.

or 

- Fork sway lock to add some functionality to make the election of fingerprint or password depending on the /etc/pam.d/swaylock pam configuration.

Elyhaka:
> Fork sway lock to add some functionality to make the election of fingerprint or password depending on the /etc/pam.d/swaylock pam configuration.

I'm facing the same behavior as the one described on the issue, and I'd like to try to implement this feature.
I imagine a flag like `--early-pam` or something like that, that would trigger the verification right away. Would this be an interesting feature to merge on swaylock once released ?

calebccff:
> > Fork sway lock to add some functionality to make the election of fingerprint or password depending on the /etc/pam.d/swaylock pam configuration.
> 
> I'm facing the same behavior as the one described in this issue, and I'd like to try to implement this feature.
> I imagine a flag like `--early-pam` or something like that, that would trigger the verification right away. Would this be an interesting feature to merge on swaylock once implemented ?

Would be sweet if you're still thinking about it, an important note is that fprintd will (and should) timeout, so I'm not sure how feasible this is.

ardevd:
Any news on this? I'm having a hard time with swaylock after adding fingerprint auth in Gnome.

calebccff:
[...]

PR #389, opened 2025-01-04
Drive interaction from authentication backend (PAM)

mgunyho:
This is related to #61, #213, #244, possibly #270.

Instead of blocking until the password is submitted, the backend process goes directly to `pam_authenticate` and displays messages and prompts to the user and reads input based on the PAM conversation. The main change is in e0775f4.

Currently the UI part is completely missing, but hopefully with this code it's easier to test various ideas. It's pretty easy to mess up the UI state, by e.g. entering incorrect fingerprints and then an incorrect password.

Some testing is also needed with more advanced scenarios like Yubikeys, I only have a fingerprint reader available.

I'm not very experienced with C and I had some trouble with casting the void pointers, I hope I didn't mess up. Please be gentle :)

The first two commits are just some additions to logging that help with debugging, let me know if I should remove them or create separate PRs.


mgunyho:
Ah and the `shadow` backend is not updated yet, that's why the build failed

mgunyho:
I have added a simple UI to show the messages from PAM, here is a quick video showing how it interacts with the typing indicator and what happens when the fingerprint validation fails.


https://github.com/user-attachments/assets/23894d40-e84f-4211-98e4-2ab175056b5b

mgunyho:
One issue with the UI currently is that if PAM is waiting for the fingerprint but you submit a password, it says "verifying" which might be misleading

comm.c
```diff
@@ -9,41 +9,87 @@
 
 static int comm[2][2] = {{-1, -1}, {-1, -1}};
 
-ssize_t read_comm_request(char **buf_ptr) {
+ssize_t write_string(int fd, const char * const *string, size_t len) {
```
mstoeckl, line 12: I don't have the time and setup to comment on the UI right now, but can advise on matters of C.

A double pointer is unnecessary when writing strings; this function and the functions which call it should use a plain `const char *string` instead.

(Using unnecessary pointers is also risky: the current `write(fd, string[offs], len - offs);` compiles fine, but actually has a bug because `string[offs]` indexes the outer pointer and not the inner one; with a `const char *string` argument, one has `write(fd, &string[offs], len - offs);` instead.)

pam.c
```diff
@@ -35,17 +33,41 @@ static int handle_conversation(int num_msg, const struct pam_message **msg,
...
+			#endif
+
+		case PAM_PROMPT_ECHO_OFF:
+			char *pw_buf = NULL;
+			ssize_t size = read_comm_prompt_response(&pw_buf);
+			if (size < 0) {
+				swaylock_log(LOG_ERROR, "Failed to read prompt response");
+				return PAM_ABORT;
+			}
+			//TODO: what to do if size == 0? in fbc5a8136187, if read_comm_prompt_response (there called read_comm_request) returned 0, it would break the while loop (i.e. auth successfully); we can't do that here
```
mstoeckl, line 55: If `size==0`, then the pipe connecting to the main `swaylock` process has disconnected because e.g. swaylock has exited (e.g. unlocked by signal, or crashed). I think the correct thing to do here would be to return `PAM_ABORT`, so that `pam_authenticate` returns `PAM_ABORT`; the current authentication attempt cannot complete.

If you furthermore want `run_pw_backend_child()` to not warn about this specific exit path, you could provide it with additional information by passing a pointer to a variable declared in `run_pw_backend_child` as  `.appdata_ptr` in `struct pam_conv conv`, and having `handle_conversation` modify the variable depending on the reason for exit.

mgunyho: Oh, you are right. I guess my thinking was that "`char *` is a string, and a string should be passed by reference instead of value". But of course it's already a reference because it's a pointer. I fixed this now in bc18cc2 (I can clean up the commit history at the end.)

I also see that `string[offs]` is wrong, but how come it works? ... Oh is it because `offs` always happens to be zero?

comm.c
mgunyho: Oh, you are right. I guess my thinking was that "`char *` is a string, and a string should be passed by reference instead of value". But of course it's already a reference because it's a pointer. I fixed this now in bc18cc2 (I can clean up the commit history at the end.)

I also see that `string[offs]` is wrong, but how come it works? ... Oh is it because `offs` always happens to be zero?

pam.c
mgunyho: Yeah, it makes sense to return `PAM_ABORT`. I think I was confused because in the [current version](https://github.com/swaywm/swaylock/blob/fbc5a8136187491e5ee3071e8a60e6593ca9b90b/pam.c#L95-L102), `size == 0` breaks the while loop and exits with `EXIT_SUCCESS`, so I thought it indicates success, but indeed the subprocess exit status doesn't necessarily say anything about success on the main process side.

mgunyho:
I've now been using this for the past month, and I have to say, it's not perfect.

It mostly works fine, but maybe once a day there is a situation where I open my laptop, and the "scan your fingerprint" message is displayed on screen, but neither scanning the fingerprint or entering the password works. This lasts for around 10 seconds after which it's possible to log in (maybe the fp reader takes a while to wake up, or something times out?), sometimes with the fingerprint but sometimes only with the password.

I suspect that somehow the PAM conversation is initiated before the laptop goes to sleep (when e.g. closing the lid), and that's when the message is displayed. Then the message is still there after opening the lid, even though the fp reader is not yet active.

I would imagine this is very confusing for a regular user - it's pretty confusing to me even though I implemented this feature and I know how it works internally! Maybe it's not worth it to merge this in its current form.

walking-octopus:
Will this PR be merged any time soon?

mgunyho:
I haven't worked on this no, apart from daily driving 60943f8 since January. 

Basically it's like I said before, the message saying "place your finger on the fingerprint reader" is sometimes shown when it's not supposed to (for example if the message is displayed before my laptop goes to sleep), which is very confusing UX. I haven't tried if it could be fixed by fiddling with how swaylock is launched by swayidle.

PR #393, opened 2025-01-28
Fix segfault due to dereferencing NULL keymap

MStarvik:
On some systems, frames may be rendered before a keymap is received, which leads to a segfault [when the keymap pointer is dereferenced](https://github.com/swaywm/swaylock/blob/fbc5a8136187491e5ee3071e8a60e6593ca9b90b/render.c#L169). This seems to happen when swaylock is started after triggering suspend, or shortly before triggering it.

On my machine, this issue can reliably be reproduced using
```
systemctl suspend; swaylock
```
or using swayidle
```
swayidle -w \
    before-sleep 'loginctl lock-session' \
    lock 'swaylock -f'
```

This PR fixes the issue by checking if the keymap pointer is NULL before dereferencing.

### DIFF
--- a/render.c
+++ b/render.c
@@ -166,19 +166,21 @@ static bool render_frame(struct swaylock_surface *surface) {
 				}
 			}
 
-			xkb_layout_index_t num_layout = xkb_keymap_num_layouts(state->xkb.keymap);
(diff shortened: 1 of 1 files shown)


emersion (approved):
LGTM, thanks!

PR #399, opened 2025-03-12
Better handle subprocess crashes

emersion:
On my machine EOF doesn't trigger POLLHUP (not sure why), instead read() returns zero. It just prints the following message in a loop:

    [comm.c:101] Failed to read pw result: Success

Better handle this case in read_comm_reply() by making a difference between a failed read() and a failed authentication.

### DIFF
--- a/comm.c
+++ b/comm.c
@@ -95,13 +95,17 @@ out:
 	return result;
 }
 
-bool read_comm_reply(void) {
-	bool result = false;
-	if (read(comm[1][0], &result, sizeof(result)) != sizeof(result)) {
-		swaylock_log_errno(LOG_ERROR, "Failed to read pw result");
-		result = false;
+bool read_comm_reply(bool *auth_success) {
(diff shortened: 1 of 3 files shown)


kennylevinsen:
> On my machine EOF doesn't trigger POLLHUP (not sure why),

Weird, the only caveat seems to be that the pipe has to have seen a writer first (slightly obscured by the use of `f_pipe` which is normally just zero):

```c
static __poll_t
pipe_poll(struct file *filp, poll_table *wait)
{
	// [ ... ]
        
	mask = 0;
	if (filp->f_mode & FMODE_READ) {
		if (!pipe_empty(head, tail))
			mask |= EPOLLIN | EPOLLRDNORM;
		if (!pipe->writers && filp->f_pipe != pipe->w_counter)
			mask |= EPOLLHUP;
	}

	// [ ... ]
}
```

Oh well. Handling EOF at the read side seems proper.

kennylevinsen (approved):

PR #402, opened 2025-03-18
Define _POSIX_C_SOURCE globally and detect supported warnings

emersion:
### Define _POSIX_C_SOURCE globally

Other Sway projects have switched to this approach.

### build: use cc.get_supported_arguments() for warnings

Let's not require the C compiler to support specific warnings,
especially if half the flags are used to turn off the warnings
anyways.

### DIFF
--- a/log.c
+++ b/log.c
@@ -1,4 +1,3 @@
-#define _POSIX_C_SOURCE 199506L
 #include <errno.h>
 #include <stdarg.h>
 #include <stdio.h>
--- a/loop.c
+++ b/loop.c
@@ -1,4 +1,3 @@
-#define _POSIX_C_SOURCE 200809L
 #include <limits.h>
 #include <string.h>
 #include <stdbool.h>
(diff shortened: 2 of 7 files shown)


kennylevinsen (approved):
LGTM

landryb:
fwiw, on OpenBSD i had to revert this/drop the define from meson.build, as having POSIX_SOURCE defined disables `__BSD_VISIBLE`, which hides the `explicit_bzero()` definition used in #325.

emersion:
#325 is not merged.

landryb:
> #325 is not merged.

i know, it stalled. I'm just saying that this PR creates issues with it :)

kennylevinsen:
> which hides the explicit_bzero() definition used in https://github.com/swaywm/swaylock/pull/325.

If the OpenBSD backend should use `password_buffer_destroy` to align with the existing backends, which would solve the issue. When an exception is needed, the file can just undef _POSIX_C_SOURCE.

landryb:
@kennylevinsen thanks, that also works. I'll try to unstall #325 someday..

PR #404, opened 2025-03-21
comm: assert that password size is positive

emersion:
If it's not, we may be reading before the buffer bounds later. (The password buffer size is aligned to the page size.)

### DIFF
--- a/comm.c
+++ b/comm.c
@@ -59,6 +59,7 @@ ssize_t read_comm_request(char **buf_ptr) {
 	if (n <= 0) {
 		return n;
 	}
+	assert(size > 0);
 
 	swaylock_log(LOG_DEBUG, "received pw check request");
 


comm.c
```diff
@@ -59,6 +59,7 @@ ssize_t read_comm_request(char **buf_ptr) {
 	if (n <= 0) {
 		return n;
 	}
+	assert(size > 0);
```
Nefsen402, line 62: This assert will never run because of the if statement right above. It will short-circuit there.

Nefsen402: Never mind, I'm too tired. It's a different variable.

Nefsen402 (approved):
~~I'm confused as to how `read_full()` is used here. It seems we're using it wrong here. It looks like `read_full()` reads into a buffer until it reaches the buffer size from a fd backwards. What we're doing here is (assuming sizeof(size_t) is 8) reading 8 bytes from the fd.~~

~~Ah, it seems we have two fds, one to hold the size, and the other that is actually the password buffer. It took me a bit of head scratching before figuring that out, I think comments should be added.~~

~~Why put the buffer size into the fd? I understand that for passwords so that other processes can't read the buffer.  But does the size matter all that much?~~

Ok, this is actually much simpler than I was making it out in my head. It's just IPC. My next question is why are we spawning child a process and doing IPC for a simple password checker?

At any rate, this looks good.

PR #405, opened 2025-03-21
main: null-terminate initial password buffer

### LINKED ISSUE #403: Swaylock 1.8.1 Crashes with Empty Password
When entering an empty password, swaylock 1.8.1 crashes and gives a red screen. The logs say:

```
swaylock: ../swaylock-1.8.1/comm.c:75: read_comm_request: Assertion `buf[size - 1] == '\0'' failed.
2025-03-20 19:51:46 - [swaylock-1.8.1/main.c:1052] Password checking subprocess crashed; exiting.
```

This happens in a nested sway, but was first observed in Niri. I wasn't experiencing this on 1.8.0. Empty passwords are useful since I use fprintd.

OS: Arch

Riolku:
Fixes #403.

### DIFF
--- a/main.c
+++ b/main.c
@@ -1155,6 +1155,7 @@ int main(int argc, char **argv) {
 	if (!state.password.buffer) {
 		return EXIT_FAILURE;
 	}
+	state.password.buffer[0] = 0;
 
 	if (pipe(sigusr_fds) != 0) {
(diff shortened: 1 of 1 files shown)


Nefsen402 (approved):
Thanks!

emersion:
Eh, I totally missed that, good catch! Probably I couldn't reproduce because of optimization levels and such.

Issue #409, opened 2025-04-23
Password checking subprocess crashed; exiting.


su-sd:
Hi,

I use swaylock to lock a sway session, then suspend the pc. After wakeup, I change to another tty, login and use `killall -USR1 swaylock`.
~3 out of 4 times this works great, but sometimes sway shows a red screen, because swaylock crashed.

```
2025-04-23 01:02:22 - [swaylock/pam.c:110] pam_authenticate failed: invalid credentials
2025-04-23 01:02:25 - [swaylock/pam.c:110] pam_authenticate failed: invalid credentials
User defined signal 1
[06:03 susd@pc:~]
 $ 2025-04-23 10:57:23 - [swaylock/main.c:1052] Password checking subprocess crashed; exiting.
``` 

I am on archlinux and my system is uptodate. I haven't found out how to reliably reproduce the issue.
I found https://github.com/swaywm/swaylock/issues/403 . Not sure, if it is related or maybe is the cause of the issue. I think the issue started to occur with `1.8.2`.

mstoeckl:
One way to reliably reproduce this is to pause the main swaylock process while SIGUSR1 is being sent, and then resume it later. For example, using the following script,
```
#!/bin/bash
swaylock -d &
sleep 1
MAIN=$(pidof swaylock | tr " " "\n" | sort | head -n 1)
kill -SIGSTOP $MAIN
sleep 0.1
# This sends SIGUSR1 to both the main swaylock process and the forked PAM handling process
killall -SIGUSR1 swaylock
sleep 0.5
kill -SIGCONT $MAIN
```
I consistently get the following logs.
```
2025-04-27 10:53:58 - [swaylock/main.c:1138] Parsing CLI Args
2025-04-27 10:53:58 - [swaylock/pam.c:93] Prepared to authorize user user
2025-04-27 10:53:59 - [swaylock/main.c:1052] Password checking subprocess crashed; exiting.
```
Replacing `killall -s SIGUSR1 swaylock` with `kill -s SIGUSR1 $MAIN` makes swaylock exit cleanly.

kennylevinsen:
See if this does the trick: https://github.com/swaywm/swaylock/pull/410

su-sd:
> Are you sending that signal to swaylock?
Yes, I use `killall -USR1 swaylock`.

Big thanks to all of you!
I really appreciate the work you do in the wayland ecosystem!

su-sd:
Is there a schedule for the next release?

PR #410, opened 2025-04-27
Ignore SIGUSR1 in the backend child process

kennylevinsen:
If a user calls `killall -USR1 swaylock`, they may end up signalling the password backend. As the password backend forked before SIGUSR1 handling was set up, it inherits the default handler and terminates.

Set the SIGUSR1 handler to SIG_IGN in the child process.

Fixes: https://github.com/swaywm/swaylock/issues/409


emersion (approved):
Thanks!

PR #412, opened 2025-05-12
pam: use getauxval(AT_SECURE) for SUID check

MaxKellermann:
Comparing effective and real uid/gid is not a proper way to check for SUID execution:

1. this does not consider file capabilities

2. this check breaks when NO_NEW_PRIVS is used as the Linux kernel resets effective ids during execve(); this means the check is false, but the process still has raised capabilities

For more details about the NO_NEW_PRIVS problem, check this post and the surrounding thread:

 https://lore.kernel.org/lkml/20250509184105.840928-1-max.kellermann@ionos.com/

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -79,6 +79,7 @@ conf_data = configuration_data()
 conf_data.set_quoted('SYSCONFDIR', get_option('prefix') / get_option('sysconfdir'))
 conf_data.set_quoted('SWAYLOCK_VERSION', version)
 conf_data.set10('HAVE_GDK_PIXBUF', gdk_pixbuf.found())
+conf_data.set('HAVE_SYS_AUXV_H', host_machine.system() == 'linux' and cc.has_header('sys/auxv.h'))
 
 subdir('include')
 
--- a/pam.c
+++ b/pam.c
@@ -10,13 +10,22 @@
 #include "password-buffer.h"
 #include "swaylock.h"
 
+#ifdef HAVE_SYS_AUXV_H
+#include <sys/auxv.h> // for getauxval()
+#endif
+
 static char *pw_buf = NULL;
 
 /**
  * Are we running with elevated privileges? (e.g. setuid)
  */
 static bool is_setid() {
+#if defined(HAVE_SYS_AUXV_H) && defined(AT_SECURE)
+	/* Linux-specific */
+	return getauxval(AT_SECURE);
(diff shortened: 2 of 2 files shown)


kennylevinsen:
Well that's annoying. So the scenario we're dealing with is:

1. swaylock is compiled with PAM support.
2. swaylock is incorrectly set as SUID.
3. A program calls `prctl(PR_SET_NO_NEW_PRIVS, 1, 0, 0, 0);`, and then `exec`'s swaylock.
4. swaylock's "you're holding it wrong" warning that SUID should not be used when the PAM backend is compiled in does not trigger.

In other words, a presumably rather unusual process start could lead to a missed warning about an erroneous system configuration.

MaxKellermann:
In your description, (2) is not relevant as NO_NEW_PRIVS ignores suid anyway. This isn't about suid-executing swaylock; this is about already being SUID before swaylock gets executed. swaylock currently handles this case and emits the warning, but this ceases to work under NO_NEW_PRIVS.

But in any case, AT_SECURE is better because it is the canonical way to detect elevated privileges and also handles file capabilities (or any other future knob for privilege elevation).

Of course, it's okay to decide to ignore both of these exotic cases. But if you do that, I wonder why this check and the warning exists at all. Anybody can hold swaylock wrong all day if they want. If you want to have the warning, it should be implemented properly.

kennylevinsen:
We are not trying to detect if we have inherited capabilities in general, we are specifically detecting that swaylock itself is SUID in a configuration where that is inappropriate.

This is done because there is a different configuration (compiled without PAM) where SUID on swaylock is expected, and we try to help users avoid mistaking those two specific configurations.

In other words, it's not a security check in that sense, which is why I'm thinking twice about pulling in non-portable code and ifdefs for it. :)

(Sway itself has a SUID check added after its SUID mode of operation was deprecated to avoid accidents of similar nature: a user mistakingly leaving SUID set after SUID drop code was removed, leading to surprise root. This was meant to be a temporary error for a number of releases, but now that I think of it that was a few releases ago and I think we forgot to remove it...)

kennylevinsen:
> I believe this check should be removed entirely. It is problematic when shadow.h does not require being root to access the current user's credentials.

Note that this is about a check in the PAM backend that warns if SUID is set. It is not about checks that _enforce_ root/suid.

PR #418, opened 2025-07-15
main: change background color to #a3a3a3

n0toose:
This is a port of the change I made to i3lock a couple of years ago, see: https://github.com/i3/i3lock/pull/300/commits

FFFFFF can cause varying degrees of uncomfortableness to users like me, especially during night time. Uncomfortableness can mean "migraines" or being woken up if the electric wiring is suboptimal, causing a desktop to wake up (the latter of which is why I opened the i3lock PR, despite the effort associated with changing defaults).

However, the choice of a color that is not solid white helps users tell whether their computer is turned off or not. Light gray (#a3a3a3) is a compromise that is bright, but not "flashbang levels" of bright.

This compromise was met after a long discussion with the i3lock developers, and, given the similarities between i3lock and swaylock, this change could be sensible for swaylock as well.

### DIFF
--- a/main.c
+++ b/main.c
@@ -397,7 +397,7 @@ static void load_image(char *arg, struct swaylock_state *state) {
 }
 
 static void set_default_colors(struct swaylock_colors *colors) {
-	colors->background = 0xFFFFFFFF;
+	colors->background = 0xA3A3A3FF;
 	colors->bs_highlight = 0xDB3300FF;
 	colors->key_highlight = 0x33DB00FF;
 	colors->caps_lock_bs_highlight = 0xDB3300FF;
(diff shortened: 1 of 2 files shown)


n0toose:
Pardon my verbosity. This is my first PR here and changing defaults can be a bit intimidating. :)

main.c
```diff
@@ -33,9 +33,9 @@ static uint32_t parse_color(const char *color) {
 
 	int len = strlen(color);
 	if (len != 6 && len != 8) {
-		swaylock_log(LOG_DEBUG, "Invalid color %s, defaulting to 0xFFFFFFFF",
+		swaylock_log(LOG_DEBUG, "Invalid color %s, defaulting to 0xA3A3A3FF",
 				color);
-		return 0xFFFFFFFF;
+		return 0xA3A3A3FF;
```
emersion, line 38: This applies to all colors, not just the background. I'd prefer to leave this as-is.

emersion (commented):
Thanks for the detailed rationale in the i3 PR, this works for me! Just one comment below.

emersion (approved):
Thanks!

PR #420, opened 2025-08-05
Make PAM optional at runtime

WhyNotHugo:
swaylock links to PAM directly, so the same binary cannot be used in scenarios where PAM is not installed.

Load PAM at runtime, to enable using the same binary when PAM is not available. Use shadow as a fallback if PAM is not available.

This allows enabling both the `shadow` and `pam` backend at compile-time, but if PAM is not available at runtime, shadow is used as a fallback. This is particularly useful for distribution packages where some users rely on PAM and others do not use PAM.

I've successfully tested this on Alpine. The same binary works with or without PAM.

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -14,10 +14,14 @@ tasks:
   - setup: |
       cd swaylock
       meson build
-  - build: |
+  - build-both: |
       cd swaylock
       ninja -C build
-  - build-no-pam: |
+  - build-pam: |
       cd swaylock
-      meson configure build -Dpam=disabled
+      meson build-pam -Dshadow=disabled
+      ninja -C build
+  - build-shadow: |
+      cd swaylock
+      meson build-shadow -Dpam=disabled
       ninja -C build
--- a/backend.c
+++ b/backend.c
@@ -1,0 +1,64 @@
+#include "swaylock.h"
+#include "log.h"
+#include "config.h"
+#include <stdlib.h>
+
+#if !HAVE_PAM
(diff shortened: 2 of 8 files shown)


WhyNotHugo:
It needs PAM at compile-time, but works without PAM at runtime.

The main benefit is that a single binary can be use both with and without PAM.

WhyNotHugo:
I've updated the PR description with the details of the final version. I've both build permutations to CI, and this is ready for review.

WhyNotHugo:
To clarify on the intention behind this: on Alpine we have a `swaylock` package where the maintainer opted to build it with PAM support. This makes the package unsuitable for setups which don't use PAM.

If the maintainer chooses to build `swaylock` _without_ PAM support, that would break for a different group of users.

By making PAM optional at runtime, the same builds are usable by anyone, regardless of whether they use PAM or not.

I'm sure other distributions where PAM is not part of the base system face a similar situation.

WhyNotHugo:
Removed extraneous commit.

emersion:
Can't distros ship two packages, one with PAM and one without?

WhyNotHugo:
Technically: yes. Two different packages can ship different builds of the same program with different features enabled. 

Users needs to know which one to install and pick the right one.

But there's often resistance to having to multiple versions of the same package (and there is in this case).

We don't really need to have the backend specified at build time and can determine it at runtime, so that seems like the natural path. Do you see any risks in this implementation?

ebenali:
like the idea

PR #421, opened 2025-08-09
fix missing include for sigaction

filko:


emersion (approved):
Thanks!

Which OS are you building swaylock on?

filko:
> Which OS are you building swaylock on?

Exherbo linux, gcc 14.3.0, glibc 2.41.

PR #423, opened 2025-08-15
check for pending callback on destroy_surface

apreiml:
Avoids segfaults when a pending surface_frame_handle_done is invoked after destruction of the associated surface.

This happened when an output was removed and the surfaces have been destroyed in between render and  surface_frame_handle_done.

### DIFF
--- a/main.c
+++ b/main.c
@@ -92,6 +92,9 @@ static void daemonize(void) {
 }
 
 static void destroy_surface(struct swaylock_surface *surface) {
+	if (surface->frame != NULL) {
+		wl_callback_destroy(surface->frame);
+	}
 	wl_list_remove(&surface->link);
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -92,6 +92,9 @@ static void daemonize(void) {
 }
 
 static void destroy_surface(struct swaylock_surface *surface) {
+	if (surface->frame != NULL) {
+		wl_callback_destroy(surface->frame);
+	};
```
emersion, line 97: Nit: extraneous semicolon on this line

emersion (commented):
LGTM apart from this!

main.c
apreiml: Oops! Fixed it.

emersion (approved):
Thanks!

PR #425, opened 2025-08-26
Wait for surface presentation before daemonizing

kennylevinsen:
We currently wait for the session to enter the "locked" state before setting the ready state and daemonizing. Lock surface configuration and rendition happens in paralle with this, and there is no guarantee that all displays have had new frames submitted at this point.

Instead, use presentation-time to wait for a single `presented` event on all surfaces, or surface removal, before continuing.


emersion:
Hm, I'm not sure I'm following… What do you mean by "that's orthogonal to whether swaylock surface content has been updated and presented"?

Issue #143, opened 2020-06-08
Add grace period for the screen locker


tpircher:
Please consider a grace period for the screen locker, similar to the `lockTimeout` setting in xscreensaver. This allows for a certain amount of time to exit the screen lock without having to type in the password.
One example where this may be useful is when watching a video or when in a conf call, when mouse and keyboard are idle for a longer period, but the screen should be visible. In that case, wiggling the mouse is much less intrusive than having to type in the password.

The xscreensaver option is documented as:

> lockTimeout (class Time)
>     If locking is enabled, this controls the length of the ''grace period'' between when the screensaver activates, and when the screen becomes locked. For example, if this is 5, and -timeout is 10, then after 10 minutes, the screen would blank. If there was user activity at 12 minutes, no password would be required to un-blank the screen. But, if there was user activity at 15 minutes or later (that is, -lock-timeout minutes after activation) then a password would be required. The default is 0, meaning that if locking is enabled, then a password will be required as soon as the screen blanks.

mortie:
This has already been discussed and rejected: https://github.com/swaywm/swaylock/issues/58, https://github.com/swaywm/swaylock/issues/50.

If you want a grace period option, you can check out my own [swaylock-effects](https://github.com/mortie/swaylock-effects) where I merged (a version of) the `--grace` option suggested by gdamjan: https://github.com/mortie/swaylock-effects/commit/c0f6b94f9f732429f23a8696578a1260c010ce13

shibe2:
> when mouse and keyboard are idle for a longer period

swaylock doesn't monitor mouse and keyboard, some other program does it. What you want is a warning when the screen is about to be locked because of inactivity. That program which monitors user activity should give you a warning before starting swaylock. Then this grace feature doesn't need to be in swaylock.

> when watching a video

Ideally, the screen shouldn't be locked at all while you are watching a video. Again, the program that monitors user activity should deal with it.

> xscreensaver option

XScreenSaver does both activity monitoring and screen locking, so naturally it has all the relevant timeout options.

532910:
> swaylock doesn't monitor mouse and keyboard

it must not monitor mouse and keyboard for grace period

> What you want is a warning when the screen is about to be locked because of inactivity.

No warning must be shown. Just a timeout before lock.

shibe2:
> No warning must be shown. Just a timeout before lock.

By "warning" I mean some kind of indication, not necessarily a warning message. What you want to happen when there is no user input:

[...]

PR #456, opened 2026-04-19
Add optional on-startup grace period; support HTML web color specification for cmdline/config

ebenali:
* add option for a grace period during startup to quickly dismiss lock without auth
    - a keystroke delivered before this grace timeout elapses will return
      user to desktop.
    - intended to be used along an idle-based desktop daemon like `swayidle`:
      ```
      # e.g: sway/config, allow 3 seconds of grace timeout
      exec swayidle -w \
         timeout 1200 'swaylock -f -g 3'
* add parsing HTML/web colors by name: `swaylock -c aqua`


emersion:
The grace period was discussed before and rejected: https://github.com/swaywm/swaylock/issues/143

I don't think the HTML color list is valuable enough to warrant its inclusion.

ebenali:
> The grace period was discussed before and rejected: https://github.com/swaywm/swaylock/issues/143
> 
> I don't think the HTML color list is valuable enough to warrant its inclusion.

well I guess that's how everybody's gonna be customizing their swaylock, which I see has been goinng on for years.. fork and compile and not use prior forks because most aren't synced up

PR #465, opened 2026-07-08
Fix stray character in password buffer after suspend/resume cycle

### LINKED ISSUE #306: Garbage characters already present in place of password after waking up from sleep mode
I have noticed it happen multiple times that after waking up from sleep, before I type my password I have to first hold backspace to clear some garbage value that is already typed there. It can be anything from 2-3 characters to 2-3 words size garbage. 

I am currently using swaylock version v1.6.11.
This is the script I am using to lock screen and I have mapped it to a keybinging in sway config file.
```
#!/bin/sh
# Times the screen off and puts it to background
swayidle \
    timeout 10 'swaymsg "output * dpms off"' \
    resume 'swaymsg "output * dpms on"' \
    timeout 60 'systemctl suspend ' &
# Locks the screen immediately
swaylock --clock --indicator --effect-scale 0.4 --effect-vignette 0.2:0.5 --effect-blur 4x2 --datestr "%a %e.%m.%Y" --timestr "%k:%M" --ignore-empty-password --show-failed-attempts --indicator-caps-lock -i ~/.config/lockscreen-wall.jpg --indicator-radius 75 --indicator-thickness=15
# Kills last background task so idle timer doesn't keep running
kill %%
```
Since I do not know how to approach debugging this issue so this is the best I can provide now, but I am happy to share any info or log that might be needed to fix this.

derconno:
This should fix https://github.com/swaywm/swaylock/issues/306

### DIFF
--- a/seat.c
+++ b/seat.c
@@ -45,6 +45,13 @@ static void keyboard_keymap(void *data, struct wl_keyboard *wl_keyboard,
 	state->xkb.state = xkb_state;
 }
 
+static void stop_repeat(struct swaylock_seat *seat) {
+	if (seat->repeat_timer) {
+		loop_remove_timer(seat->state->eventloop, seat->repeat_timer);
+		seat->repeat_timer = NULL;
+	}
+}
+
 static void keyboard_enter(void *data, struct wl_keyboard *wl_keyboard,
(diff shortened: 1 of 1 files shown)


emersion (approved):
Good catch, thanks!

